Issue #32, opened 2021-08-25
[Feature] Ability to exclude directory


adamzimmermann:
This is a shameless feature request that would solve an issue I've stumbled upon as I'm integrating this tool with more repositories. Your tool has been working wonderfully with everything it promises to do and has been a big help so far.

Some context:
- I have a yml file that contains the single canonical list of custom code paths in a codebase that tells `phpcs` and `twig-cs-fixer` the paths where there is custom code.
  - https://github.com/ChromaticHQ/chq-robo/blob/2.x/src/Robo/Plugin/Commands/CICommands.php#L54
  - https://github.com/ChromaticHQ/chq-robo/blob/2.x/src/Robo/Plugin/Commands/CICommands.php#L104-L109
- Within one of those paths is a theme that passes linting normally. But after running `yarn build` the `node_modules` folder has some Twig files that the linter catches and flags as issues.

Could we add the ability for `twig-cs-fixer` to support an `--ignore` flag the way `phpcs` does?

e.g. https://github.com/ChromaticHQ/chq-robo/blob/2.x/src/Robo/Plugin/Commands/CICommands.php#L106

I realize that I could simply separate the list of custom code paths and only point `twig-cs-fixer` at the `templates` directory in my theme. However, then there is the need to maintain two lists. I'm guessing others might find this feature useful for other purposes as well.

I assume the logic would live somewhere in here?
- https://github.com/VincentLanglet/Twig-CS-Fixer/blob/313b82affe4c3c9d2609d43d93dc9b9d6163799c/src/File/Finder.php#L43

If so, I can take a shot at contributing something, but wanted to get your take first.

### Expected behavior

The ability to ignore selected directories.

### Actual behavior

Unable to do so.

PR #33, opened 2021-08-26
Add --exclude Flag and use Symfony Finder

adamzimmermann:
## Description

This is _not_ working currently, but I wanted to get it up here to see if I'm at least roughly in line with the approach you were thinking.

This PR is attempting to partially revert these changes:
- https://github.com/VincentLanglet/symfony-custom-coding-standard/pull/114/files

One thing I'm not 100% clear on is why we want to keep `TwigCsFixer\File\Finder.php` or add configurability around that. Why would we not just always use the Symfony Finder? Or perhaps I'm misunderstanding your intent?

## Motivation
- #32 

## Remaining Tasks

- [x] Decide if we need to support both Finders
- [x] Get it to actually work.
- [ ] Update tests if applicable?
- [ ] Update the README with documentation about the new option.


VincentLanglet:
@adamzimmermann This is what I had in mind https://github.com/VincentLanglet/Twig-CS-Fixer/commit/06089cf61e715a577d486bdfd90356c3b554da1c

The `path()` is not the right way to filter path when provided, because if I run on the path `foo`, and there is a path `bar/foo/` this one will be kept. Instead you have to basically define a new Finder.
For this reason, the `in('./')` call must be removed from the `Finder` constructor, and added later, only if no `in` call was made.
Since resolving the Finder with the path provided is not a simply task, I prefer to not add an `exclude` option and only rely on the configuration.
I also fix the lint errors.

There is still test to fix/add.

adamzimmermann:
@VincentLanglet I'll admit that I'm not fully understanding the approach you are describing, but it looks like you are well on your way to a refactor in your [new PR](https://github.com/VincentLanglet/Twig-CS-Fixer/pull/35)!

I need to focus on some other tasks this week anyways, so I might just let you take finish the code changes, but I'm happy to help test this when it's ready!

PR #35, opened 2021-08-29
Use finder to configure paths

VincentLanglet:


VincentLanglet:
@adamzimmermann This is supposed to be working if you can make a try for your use case.

adamzimmermann:
Sounds good. I'll try to make some time to test this in the next day or so.

adamzimmermann:
@VincentLanglet I just got around to testing this and I'm not fully following your recommendations.

I'm not sure what you are recommending here:
- https://github.com/VincentLanglet/Twig-CS-Fixer/pull/33#issuecomment-907847448

...and how to turn that into an actionable change to my repo to test your latest changes.

I have my repo using the branch from this PR and have a `.twig-cs-fixer.php` file setup as follows:

```
<?php

use TwigCsFixer\File\Finder;

$finder = new Finder();

// This is where I'm unsure what you are suggesting I should do, or if this is even the right spot.
// Or should I be using the `--paths` flag?

$config = new TwigCsFixer\Config\Config();
$config->setFinder($finder);

return $config;

```

Suppose I have this use case:

`web/themes/mytheme` is a directory I want to lint, but `web/themes/mytheme/node_modules` is a directory I do not want to lint. How would I do that?

I could say that I only want to lint `web/themes/mytheme/templates` to help solve this, but I would like to avoid that as I would prefer to point to the whole theme (Or perhaps this more specific subdirectory isn't an option for someone else). As I'm trying to share config that has a list of all custom code paths with this linting tool and some other PHP linting tools.

For more background: We require `chromatichq/chq-robo` in all of our repos and it brings in all of the linting tooling and simply reads from a yml file each repo provides that lists all custom code paths with no other setup needed.
- https://github.com/ChromaticHQ/chq-robo/blob/2.x/src/Robo/Plugin/Commands/CICommands.php#L108

Let me know if any of that does not make sense.

I'm excited to test this out as soon as I understand how to fully!

adamzimmermann:
@VincentLanglet ah it's that simple. I thought I needed to do something with `paths()`. 

I'll test this again shortly. Thank you!

adamzimmermann:
It's "working" sort of, but I'm not able to pass exclusions as you suggested. At least with my understanding of what you suggested. Correct me if I'm wrong though.



---

### This works:

Screen Shot 2021-09-02 at 10 47 09 AM
⬇️ 
Screen Shot 2021-09-02 at 10 46 58 AM

---

### This is still broken:
Screen Shot 2021-09-02 at 10 47 26 AM

Screen Shot 2021-09-02 at 10 47 47 AM

VincentLanglet:
> It's "working" sort of, but I'm not able to pass exclusions as you suggested. At least with my understanding of what you suggested. Correct me if I'm wrong though.

This is one limitation of the `Finder` and some of his methods.

The `exclude` path is relative to the paths passed.

With `exclude('vendor')`, when you run `lint some/path/`, it will exclude `some/path/node_modules`
But if you run `lint`, it will exclude `node_modules`.

With `exclude('some/path/vendor')`, when you run `lint some/path/`, it will exclude `some/path/some/path/node_modules`
But if you run `lint`, it will exclude `some/path/node_modules`.

So a better solution for you might be
```
$finder->notPath('node_modules')
```
to remove every path with node_modules in it.

Anyway, that's more an issue about how to use the Finder. The feature seems to work properly as expected ; I'll merge it soon.

adamzimmermann:
@VincentLanglet yep, that all lines up with what I saw when I did my initial draft of the PR in my fork, which is why I was confused by your [suggestion](https://github.com/VincentLanglet/Twig-CS-Fixer/pull/35#issuecomment-911404293).

---

 > Anyway, that's more an issue about how to use the Finder. The feature seems to work properly as expected ; I'll merge it soon.

Agreed. I should be able to sort out any issues with Finder and how it works with full control of the issue on my end with my custom `Finder`. 🎉 

Thanks for getting this integrated!

PR #66, opened 2022-10-27
adds a caching system based on PHP-CS-Fixer's

JoolsMcFly:
Adds a caching system heavily based on PHP-CS-Fixer's

Create a file named `.twig-cs-fixer.php` at the root of your project and add the following content:
```php
<?php

use TwigCsFixer\Config\Config;

$config = new Config();
$config->setCacheFile('/tmp/cache-file');

return $config;
```


src/Cache/Cache.php
```diff
@@ -0,0 +1,127 @@
...
+    public function toJson(): string
+    {
+        $json = json_encode([
+            'php'     => $this->getSignature()->getPhpVersion(),
+            'version' => $this->getSignature()->getFixerVersion(),
+            'sniffs'  => array_keys($this->getSignature()->getRuleset()->getSniffs()),
+            'hashes'  => $this->hashes,
+        ]);
+
+        if (\JSON_ERROR_NONE !== json_last_error()) {
```
VincentLanglet, line 69: I prefer to use the flag `\JSON_THROW_ON_ERROR` when encoding. (We can still catch JSONException to improve the error message if needed)

src/Cache/Directory.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+declare(strict_types=1);
+
+namespace TwigCsFixer\Cache;
+
+/**
+ * @internal
+ */
```
VincentLanglet, line 9:
```suggestion
```

src/Cache/FileHandlerInterface.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+declare(strict_types=1);
+
+namespace TwigCsFixer\Cache;
+
+/**
+ * @internal
+ */
```
VincentLanglet, line 9:
```suggestion
```

src/Runner/Linter.php
```diff
@@ -82,13 +89,18 @@ private function fix(iterable $finder, Ruleset $ruleset): void
 
         foreach ($finder as $file) {
             $filePath = $file->getPathname();
-            $success = $fixer->fixFile($filePath);
+            $success = true;
+            if ($this->cacheManager->needFixing($filePath, file_get_contents($filePath))) {
+                $success = $fixer->fixFile($filePath);
+            }
```
VincentLanglet, line 95:
```suggestion
            if (!$this->cacheManager->needFixing($filePath, file_get_contents($filePath))) {
                continue;
            }
            $success = $fixer->fixFile($filePath);
```
no ?

src/ToolInfo/ToolInfoInterface.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace TwigCsFixer\ToolInfo;
+
+/**
+ * @internal
+ */
```
VincentLanglet, line 9:
```suggestion
```

src/ToolInfo/ToolInfo.php
```diff
@@ -0,0 +1,52 @@
...
+            $packages = $composerInstalled['packages'] ?? $composerInstalled;
+
+            foreach ($packages as $package) {
+                if (self::COMPOSER_PACKAGE_NAME === $package['name']) {
+                    $this->composerInstallationDetails = $package;
+
+                    break;
+                }
+            }
+        }
```
VincentLanglet, line 30: I think there is already something provided by Composer.
Like `InstalledVersions::getVersion()`.

So ToolInfo and ToolInfoInterface shouldn't be needed.

VincentLanglet (changes requested):

src/Runner/Linter.php
JoolsMcFly: Yes. I went a bit further as phpstan complained about possible `false` value being passed to `needFixing` param #2.

```php
            $contents = file_get_contents($filePath);
            if (false === $contents || !$this->cacheManager->needFixing($filePath, $contents)) {
                continue;
            }

            if (!$fixer->fixFile($filePath)) {
                throw new Exception(sprintf('Cannot fix the file "%s".', $filePath));
            }
```

I got rid of `$success` in the process.

src/ToolInfo/ToolInfo.php
```diff
@@ -0,0 +1,52 @@
...
+declare(strict_types=1);
+
+namespace TwigCsFixer\ToolInfo;
+
+/**
+ * Obtain information about using version of tool.
+ */
+final class ToolInfo implements ToolInfoInterface
+{
+    public const COMPOSER_PACKAGE_NAME = 'k10r/twig-cs-fixer';
```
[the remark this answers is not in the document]
JoolsMcFly: Just a brain fart. :)

src/Cache/FileCacheManager.php
```diff
@@ -0,0 +1,106 @@
...
+    private SignatureInterface $signature;
+
+    private CacheInterface $cache;
+
+    private ?DirectoryInterface $cacheDirectory;
+
+    public function __construct(
+        FileHandlerInterface $handler,
+        SignatureInterface $signature,
+        ?DirectoryInterface $cacheDirectory = null
```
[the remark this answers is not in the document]
JoolsMcFly: We can!

psalm.xml
```diff
@@ -13,6 +13,7 @@
        inferPropertyTypesFromConstructor="false"
        findUnusedPsalmSuppress="true"
        cacheDirectory="/tmp/psalm"
+       allowStringToStandInForClass="true"
```
VincentLanglet, line 16: Why do you need this ? We're not supposed to

src/Cache/Cache.php
```diff
@@ -117,23 +119,34 @@ public static function fromJson(string $json): self
 
         $ruleSet = new RuleSet();
         if (\is_array($data['sniffs'])) {
+            /** @psalm-var class-string $sniffName */
             foreach ($data['sniffs'] as $sniffName) {
-                $sniff = new $sniffName();
-                if ($sniff instanceof SniffInterface) {
-                    $ruleSet->addSniff($sniff);
+                try {
```
VincentLanglet, line 124: I'll take a look this We if we can improve this.

JoolsMcFly: I've committed something better as suggested by someone at Psalm.

psalm.xml
JoolsMcFly: Psalm was complaining and it fixed my issue at some point but now it's been refactored I don't need it anymore. It's now removed.

src/Cache/Cache.php
VincentLanglet: This is really better indeed. thanks.

src/Cache/Cache.php
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    public function has(string $file): bool
+    {
+        return \array_key_exists($file, $this->hashes);
+    }
+
+    /**
+     * @throws LogicException
+     */
```
VincentLanglet, line 40: I don't think I added `@throws` for logicException since it's not supposed to be caught.

src/Cache/Cache.php
```diff
@@ -0,0 +1,147 @@
...
+    public function get(string $file): int
+    {
+        if (!$this->has($file)) {
+            throw new LogicException('You should call the \'has\' method prior to calling \'get\'');
```
VincentLanglet, line 44: I would prefer something like
```suggestion
            throw new LogicException(sprintf('The file %s is not cached', $file));
```

src/Cache/Cache.php
```diff
@@ -0,0 +1,147 @@
...
+                'php'     => $this->getSignature()->getPhpVersion(),
+                'version' => $this->getSignature()->getFixerVersion(),
+                'sniffs'  => array_keys($this->getSignature()->getRuleset()->getSniffs()),
+                'hashes'  => $this->hashes,
+            ], \JSON_THROW_ON_ERROR);
+        } catch (JsonException $e) {
+            $error = sprintf('Cannot encode cache signature to JSON, error: "%s".', $e->getMessage());
+            if (\in_array($e->getCode(), [\JSON_ERROR_UTF8, \JSON_ERROR_UTF16], true)) {
+                $error .= ' If you have non-UTF8 or non-UTF16 chars in your signature, like in license for `header_comment`, consider enabling `ext-mbstring` or install `symfony/polyfill-mbstring`.';
+            }
```
VincentLanglet, line 76:
```suggestion

            if (\in_array($e->getCode(), [\JSON_ERROR_UTF8, \JSON_ERROR_UTF16], true)) {
                $error .= ' If you have non-UTF8 or non-UTF16 chars in your signature, like in license for `header_comment`, consider enabling `ext-mbstring` or install `symfony/polyfill-mbstring`.';
            }

```

src/Cache/Cache.php
```diff
@@ -0,0 +1,147 @@
...
+
+        $ruleSet = new RuleSet();
+        if (\is_array($data['sniffs'])) {
+            /** @psalm-var class-string $sniffName */
+            foreach ($data['sniffs'] as $sniffName) {
+                if (!is_a($sniffName, SniffInterface::class, true)) {
+                    continue;
+                }
+                $ruleSet->addSniff(new $sniffName());
+            }
```
VincentLanglet, line 126: I think you can do

```suggestion
            foreach ($data['sniffs'] as $sniffName) {
                if (!class_exists($sniffName)) {
                    throw new \InvalidArgumentException(...);
                }
                if (!is_a($sniffName, SniffInterface::class, true)) {
                    throw new \InvalidArgumentException(...);
                }
                $ruleSet->addSniff(new $sniffName());
            }
```

- You won't need `/** @psalm-var class-string $sniffName */`
- I prefer exception over continue, since it's not supposed to happen ; right ?

src/Cache/Cache.php
```diff
@@ -0,0 +1,147 @@
...
+
+        if (\count($missingKeys) > 0) {
+            throw new InvalidArgumentException(sprintf(
+                'JSON data is missing keys "%s"',
+                implode('", "', $missingKeys)
+            ));
+        }
+
+        $ruleSet = new RuleSet();
+        if (\is_array($data['sniffs'])) {
```
VincentLanglet, line 119:
```
if (!isset($data['sniffs']) || !is_array($data['sniffs']) {
    throw new \InvalidArgumentException(...);
}
```

src/Cache/Cache.php
```diff
@@ -0,0 +1,147 @@
...
+                if (!is_a($sniffName, SniffInterface::class, true)) {
+                    continue;
+                }
+                $ruleSet->addSniff(new $sniffName());
+            }
+        }
+
+        $signature = new Signature(
+            (string) $data['php'],
+            (string) $data['version'],
```
VincentLanglet, line 131: In the same way, I will prefer to not cast silently this
```
if (!isset($data['php']) || !is_string($data['php']) {
     throw new \invalidArgumentException(...);
}
```
etc

src/Cache/Cache.php
```diff
@@ -0,0 +1,147 @@
...
+            $ruleSet,
+        );
+
+        $cache = new self($signature);
+
+        /**
+         * @var string $file
+         * @var int    $hash
+         */
+        foreach ((array) $data['hashes'] as $file => $hash) {
```
VincentLanglet, line 141: In the same way, I prefer
- a `is_array` check for data['hashes']
- a `is_string` check for $file rather than phpdoc.

src/Cache/Cache.php
```diff
@@ -0,0 +1,147 @@
...
+    public function get(string $file): int
+    {
+        if (!$this->has($file)) {
+            throw new LogicException('You should call the \'has\' method prior to calling \'get\'');
+        }
+
+        return $this->hashes[$file];
+    }
+
+    public function set(string $file, int $hash): void
```
VincentLanglet, line 50: I think we could support `int|string $hash` no ?

src/Config/Config.php
```diff
@@ -20,12 +29,22 @@ final class Config
 
     private Finder $finder;
 
-    public function __construct(string $name = 'Default')
+    private ?string $cacheFile = null;
```
VincentLanglet, line 32: Should we always provide cache by default with a default cache file ?

Phpcsfixer use `.php-cs-fixer.cache` by default.
We could use `.twig-cs-fixer.cache`.

This way, even if people are not aware of this new feature (or doesn't want to have a custom conf) they will enjoy the caching feature.

src/Runner/Linter.php
```diff
@@ -82,13 +89,18 @@ private function fix(iterable $finder, Ruleset $ruleset): void
 
         foreach ($finder as $file) {
             $filePath = $file->getPathname();
-            $success = $fixer->fixFile($filePath);
+            $contents = file_get_contents($filePath);
+            if (false === $contents || !$this->cacheManager->needFixing($filePath, $contents)) {
```
VincentLanglet, line 93: if `false === $contents` I think an Exception should be thrown.

VincentLanglet (commented):
This is already done, we'll just need more tests. Do you want help about it ?

src/Cache/Cache.php
```diff
@@ -0,0 +1,147 @@
...
+        return $json;
+    }
+
+    /**
+     * @throws InvalidArgumentException
+     */
+    public static function fromJson(string $json): self
+    {
+        try {
+            $data = json_decode($json, true, 512, \JSON_THROW_ON_ERROR);
```
VincentLanglet, line 89: We need to add `ext-json` to the composer.json

src/Cache/FileHandler.php
```diff
@@ -0,0 +1,91 @@
...
+
+        return Cache::fromJson($content);
+    }
+
+    /**
+     * @throws IOException
+     */
+    public function write(CacheInterface $cache): void
+    {
+        $content = $cache->toJson();
```
VincentLanglet, line 42: There is something wrong here.

We're calling `$cache->toJson();`, using the implementation of toJson of the CacheInterface which can be custom.

But we're using `Cache::fromJson($content)` which is our implementation so might not work with a custom implementation.

Instead of providing a `toJson` method in the CacheInterface, we would need to expose `getHashes` or `getAll` (or a better name) so we might create a service with the toJson/fromJson method (or implement them directly in the FileHandler).

src/Cache/FileCacheManager.php
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    private function writeCache(): void
+    {
+        $this->handler->write($this->cache);
+    }
+
+    private function calcHash(string $content): int
+    {
+        return crc32($content);
```
VincentLanglet, line 102: Seems like md5 is better
https://github.com/FriendsOfPHP/PHP-CS-Fixer/commit/97f5a4494f507cd42ccf28925bee3358fabd3653

So Hash will be string now.

VincentLanglet (changes requested):

src/Cache/Cache.php
JoolsMcFly: Not sure what changed here, contents are identical or am I missing something?

src/Cache/Cache.php
JoolsMcFly: I added checks and got rid of `psalm-var`

src/Cache/Cache.php
JoolsMcFly: I am now throwing an exception but `isset` is not needed as we make sure all keys are there (`$missingKeys`).

src/Runner/Linter.php
JoolsMcFly: Done deal

src/Config/Config.php
```diff
@@ -20,12 +29,22 @@ final class Config
 
     private Finder $finder;
 
-    public function __construct(string $name = 'Default')
+    private ?string $cacheFile = '.twig-cs-fixer.cache';
```
VincentLanglet, line 32:
```suggestion
    private string $cacheFile = '.twig-cs-fixer.cache';
```
Since we set a default value

src/Config/Config.php
```diff
@@ -62,4 +81,62 @@ public function setFinder(Finder $finder): self
 
         return $this;
     }
+
+    public function getCacheManager(): CacheManagerInterface
+    {
+        if (null === $this->cacheManager) {
+            $cacheFile = $this->getCacheFile();
+
+            if (null === $cacheFile) {
```
VincentLanglet, line 90: This shouldn't be possible.

src/Config/Config.php
```diff
@@ -62,4 +81,62 @@ public function setFinder(Finder $finder): self
...
+                    : $this->cwd.\DIRECTORY_SEPARATOR.$path;
+            }
+
+            $this->directory = new Directory(\dirname($absolutePath));
+        }
+
+        return $this->directory;
+    }
+
+    public function getCacheFile(): ?string
```
VincentLanglet, line 128:
```suggestion
    public function getCacheFile(): string
```

VincentLanglet (changes requested):

src/Cache/FileHandler.php
```diff
@@ -0,0 +1,208 @@
...
+                null,
+                $this->file
+            );
+        }
+    }
+
+    /**
+     * @throws InvalidArgumentException
+     */
+    private function fromJson(string $json): Cache
```
VincentLanglet, line 99: Looking at how long are those methods `fromJson` and `toJson`, I think we should create a class with those two methods.

Something like a CacheSerializer with two static methods (no need for an interface then).
It will be easier to tests those methods.

JoolsMcFly: I thought so too so I created CacheEncoder and added basic tests.

VincentLanglet:
Thanks

PR #90, opened 2022-11-23
Use vendor/bin/twig-cs-fixer in README

alexander-schranz:
I replaced `bin/twig-cs-fixer` with `vendor/bin/twig-cs-fixer` as that is today more common. I also added empty spaces around code blocks as that is required for some markdown viewers sadly not remember which one it was where that problem popped up.

PS: really great project 👍

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ composer require --dev vincentlanglet/twig-cs-fixer
 ```
 
 Then, use it!
+
 ```bash
 vendor/bin/twig-cs-fixer lint /path/to/code
 vendor/bin/twig-cs-fixer lint --fix /path/to/code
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -13,14 +13,15 @@
 This standard can be installed with [Composer](https://getcomposer.org/).
 
 Add the coding standard as a dependency of your project
+
 ```bash
 composer require --dev vincentlanglet/twig-cs-fixer
 ```
 
 Then, use it!
 ```bash
```
VincentLanglet, line 22: Does it require an empty line before this one too then ?

VincentLanglet:
> PS: really great project 👍

Thanks a lot

VincentLanglet (approved):

VincentLanglet:
Thanks for the PR

PR #99, opened 2022-12-05
Rework linter

VincentLanglet:


VincentLanglet:
Hi @JoolsMcFly, while trying to refactor the linter ; I just discovered there was a bug with the cache.

We were caching every fixed files, even if there was still error on those.
One difference with PHP-Cs-Fixer is the fact that I allow people to implements fixable errors and non-fixable errors.
If you have a non-fixable errors, I shouldn't cache the file, or the next time you run the linter, the file will be skipped and you won't know the file has an error.

I tried to fix this and made a multiple changes to the linter (with more tests).
I hope I didn't break any stuff.

JoolsMcFly:
Bit late to the game but indeed, any exception thrown by `fixFile` was caught but the script could continue and add current file to cache. Looks OK now.

Issue #47, opened 2022-01-13
html & twig cs fixer


rafipiccolo:
i dont know if i could help or something, but i saw this [issue](https://github.com/VincentLanglet/Twig-CS-Fixer/issues/43)
and i was kind of sad to see there was no solution.
I'm curious to know which html formatter you actually use, that would be compatible with twig.
i'm a bit tired of new / old developers pushing untidy code, having missing/mismatch closing tags because bad indent, etc.
it's time to have a real html+twig formatter or ditch twig for good.

I have to say that vscode's default html formater destroys twig templates (for exemple "#}" => "# }" which kills rendering).

i just want something that works, any specific html cs/twig cs woud be acceptable, but here is my favorite :
indent all html tags (one per line), then indent twig correctly (one per line), and when the line is too large for a specific tag we split attributes line by lines.

```
<div style="{% if xx %}fdsfsdfsdf{% endif %}">
    {% if xxx %}
        <a
            href="xxxxxxxxxxxx"
            style="xxxxxxxxxxxxxxx"
        >
            <img src="" />
        </a>
    {% endif %}
</div>
```

everything is f**** readable now :) 👍

VincentLanglet:
> I'm curious to know which html formatter you actually use, that would be compatible with twig.

Unfortunately, I don't have one. I have currently nothing to endorse the following format
```
<div style="{% if xx %}fdsfsdfsdf{% endif %}">
    {% if xxx %}
        <a
            href="xxxxxxxxxxxx"
            style="xxxxxxxxxxxxxxx"
        >
            <img src="" />
        </a>
    {% endif %}
</div>
```
The token parser see the first line of this code as
```
TEXT SPACE TEXT {% SPACE BLOCK_TAG SPACE NAME SPACE %} TEXT {% SPACE BLOCK_TAG SPACE %} TEXT EOL
```
I have no way to know when a `div` start and when it ends. 

In https://twig.symfony.com/doc/2.x/coding_standards.html there is the following standard:
- Indent your code inside tags (use the same indentation as the one used for the target language of the rendered template):
```
{% block foo %}
    {% if true %}
        true
    {% endif %}
{% endblock %}
```
But currently, nothing implemented to do something about it. And I think it will be really hard because it require to understand the block tag (`{% block foo %}` would require to indent, when `{% do foo %}` doesn't) and the html.

PR #117, opened 2023-01-30
Add indent sniff

VincentLanglet:


VincentLanglet:
@alexander-schranz Can you confirm it works for you and does what you were looking for ?

alexander-schranz:
@VincentLanglet First thank you for working on this 🙏 .

It seems correctly detect `tabs` but example false indention of spaces is not detected e.g.:

```twig
{% extends '@AvodiExtranet/base.html.twig' %}

{% block menu %}
   {% endblock %} {# 3 spaces #}

{% block main %}
  {% endblock %} {# 2 spaces #}

{% block footer %}
 {% endblock %} {# 1 spaces #}
```

Not sure if that would also be easy possible 🤔

VincentLanglet:
> It seems correctly detect `tabs` but example false indention of spaces is not detected e.g.:
> 
> ```
> {% extends 'base.html.twig' %}
> 
> {% block menu %}
>    {% endblock %} {# 3 spaces #}
> 
> {% block main %}
>   {% endblock %} {# 2 spaces #}
> 
> {% block footer %}
>  {% endblock %} {# 1 spaces #}
> ```
> 
> Not sure if that would also be easy possible 🤔

To me that's a totally different rule:
- This one is converting tab to space
- You're looking for a rule which fix space indent

Also, in the example
```
{% set foo = bar %} {# 1 spaces #}
 {% block footer %}
```
I cannot say if we want to fix this to 0 or 4 spaces.
And we'll soon end up with non-solvable issue like https://github.com/VincentLanglet/Twig-CS-Fixer/issues/47 ^^'

alexander-schranz:
Then looks fine for me :)

Understand a whole indention fix is really hard, what maybe could be possible check if a `{% end.. %}` tag is on the same indention level or same line. But as you said that would be a different rule 👍

alexander-schranz:
@VincentLanglet Thank you and congratulation for 1.0.0 🎉

VincentLanglet:
> @VincentLanglet Thank you and congratulation for 1.0.0 🎉

Thanks.
It's been years I use this without issues, so it was time for a stable version ^^

PR #119, opened 2023-02-28
Add require with only sniff

### LINKED ISSUE #118: Request: Force "include with ... only"
Hi, I stumbled upon this package while searching for a way to force `with ... only`.

Is it possible to somehow enforce this behavior?  
I'm wiling to write the sniffer myself, however I've never written one before so could use some guidance.

### Expected behavior

```diff
-{% include 'template.html' with {'foo': 'bar'} %}
+{% include 'template.html' with {'foo': 'bar'} only %}
```

### Reason

At my workplace we have a lot of templates and it becomes unclear where a used variable comes from.  
By forcing ourselves to use `with ... only` we hope to always leave a trail.

RobinHoutevelts:
Fixes #118 

This PR adds a "RequireWithOnly" sniff.

At my workplace we have a lot of templates and it becomes unclear where a used variable comes from.
By forcing ourselves to use `with ... only` we hope to always leave a trail.

If you use this in a Rule, be sure to mark it risky.
We plan to only use this in greenfield projects from the start.

```diff
-{% include 'template.html' with {'foo': 'bar'} %}
+{% include 'template.html' with {'foo': 'bar'} only %}
```

### DIFF
--- a/src/Sniff/RequireWithOnlySniff.php
+++ b/src/Sniff/RequireWithOnlySniff.php
@@ -1,0 +1,92 @@
+<?php
+
+declare(strict_types=1);
+
+namespace TwigCsFixer\Sniff;
+
+use TwigCsFixer\Token\Token;
+use Webmozart\Assert\Assert;
+
+/**
+ * Ensure includes that pass additional variables always use "with ... only".
+ */
+final class RequireWithOnlySniff extends AbstractSniff
+{
+    protected function process(int $tokenPosition, array $tokens): void
+    {
+        $token = $tokens[$tokenPosition];
+
+        // If {% is found
+        if (!$this->isTokenMatching($token, Token::BLOCK_START_TYPE)) {
+            return;
+        }
+
+        // Find the next %}
+        $endTokenPosition = $this->findNext(Token::BLOCK_END_TYPE, $tokens, $tokenPosition);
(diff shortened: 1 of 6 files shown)


VincentLanglet (commented):
Thanks for the PR.

I would say it's globally ok but I think there is an issue with findNextWithValue/findPreviousWithValue.

I think the Token::Name_Type is used for variable too, so with something like
```
{% include with %}
{% include foo with {foo: only} %}
```
For the first line, you'll think there is a with.
For the second line, you'll think there is a only.

We need to add more test to confirm this.

For the only tag, maybe it's better to do `findPrevious(Token::EmptyTokens, $endTokenPosition, true)`, it will returns the previous non empty token. Then you'll check if it's a NAME_TYPE with `only` value.

For the with, I'm no sure...

RobinHoutevelts:
Thanks for the feedback! I've put this back to DRAFT.

I also have to make sure this works with the include _function_ 🙈 
https://twig.symfony.com/doc/2.x/functions/include.html

I'll add the extra test cases. Good catch

VincentLanglet:
Tell me if you encounter any issue @RobinHoutevelts

VincentLanglet:
Any issue on this @RobinHoutevelts ?

PR #134, opened 2023-10-06
Make twig-bridge optional

VincentLanglet:


MauricioFauth (approved):

composer.json
```diff
@@ -35,9 +34,10 @@
         "phpunit/phpunit": "^9.5.26 || ^10.0.9",
         "psalm/plugin-phpunit": "^0.18.4",
         "psalm/plugin-symfony": "^5.0.0",
-        "rector/rector": "^0.17.0",
+        "rector/rector": "^0.18.0",
         "symfony/filesystem": "^5.4 || ^6.0",
         "symfony/process": "^5.4 || ^6.0",
+        "symfony/twig-bridge": "^5.4 || ^6.0",
```
MauricioFauth, line 40: I can also add this to the `suggest` section of `composer.json`.

dpi:
Just leaving a note for searchers.

We began getting errors like, in the context of Drupal projects:

> File is invalid: Unexpected "trans" tag (expecting closing tag for the "if" tag defined near line 66).
> File is invalid: Unknown "trans" tag

Since the 1.4.0 release.

Adding `symfony/twig-bridge` back as a dev dependency seems to have resolved it.

VincentLanglet:
> Just leaving a note for searchers.
> 
> We began getting errors like, in the context of Drupal projects:
> 
> > File is invalid: Unexpected "trans" tag (expecting closing tag for the "if" tag defined near line 66).
> > File is invalid: Unknown "trans" tag
> 
> Since the 1.4.0 release.
> 
> Adding `symfony/twig-bridge` back as a dev dependency seems to have resolved it.

That's kinda expected indeed.
Symfony/twig-bridge is not automatically installed anymore.

I wasn't expecting any BC break because if you're using the trans tag, you would already need the bridge to having your project working. Seems like it's not the case @dpi, how does your code work if you don't install `symfony/twig-bridge` ?
Is there something specific to drupal which understand the tag ?

dpi:
@VincentLanglet in the lock change that updated to 1.4.0, I could see this dep was removed. So nothing in Drupal or our deps relies on this. I don't know precisely what the cause is. But I imagine its because Symfony [Bridge?] implements a `trans` and Drupal does also with its own implementation. I guess if other frameworks have a `trans` tag they will also get this problem.

Looks like Drupal's is in https://github.com/drupal/drupal/blob/11.x/core/lib/Drupal/Core/Template/TwigExtension.php#L121

It doesnt seem trivial to just bring in this class into a `.twig-cs-fixer.php`, so I opted a quick fix to re-add. Do you have any suggestions about resolving `.twig-cs-fixer.php` + above `TwigExtension` class?

VincentLanglet:
> Drupal does also with its own implementation. I guess if other frameworks have a `trans` tag they will also get this problem.
> 
> Looks like Drupal's is in [drupal/drupal@`11.x`/core/lib/Drupal/Core/Template/TwigExtension.php#L121](https://github.com/drupal/drupal/blob/11.x/core/lib/Drupal/Core/Template/TwigExtension.php?rgh-link-date=2023-10-11T06%3A25%3A42Z#L121)

Thanks for the link.
The issue is more about the token parser, so https://github.com/drupal/drupal/blob/11.x/core/lib/Drupal/Core/Template/TwigTransTokenParser.php#L25

> It doesnt seem trivial to just bring in this class into a `.twig-cs-fixer.php`, so I opted a quick fix to re-add this bulky dev dep. Do you have any suggestions about resolving `.twig-cs-fixer.php` + above `TwigExtension` class?

You could try to add directly the token parser like https://github.com/VincentLanglet/Twig-CS-Fixer?tab=readme-ov-file#token-parser
and write:
```
<?php

$config = new TwigCsFixer\Config\Config();
$config->addTokenParser(new Drupal\Core\Template\TwigTransTokenParser());

return $config;
```

dpi:
Great! that works. Thanks @VincentLanglet

PR #136, opened 2023-10-12
Handle TwigComponent `props` tag

smnandre:
This PR add compatibility with the [`props`](https://symfony.com/bundles/ux-twig-component/current/index.html#anonymous-components-1) tag introduced in Twig Component 2.11, which allows to define properties for anonymous components.

```twig
{# templates/components/Button.html.twig #}
{% props icon = null, label %}

<button>
    {{ label }}
    {% if icon %}
        <span class="fa-solid fa-{{ icon }}"></span>
    {% endif %}
</button>
```


VincentLanglet:
Thanks

PR #149, opened 2023-11-22
Added checkstyle reporting.

j-barnes:
Thanks for setting up the reporting functionality, that was quick!

Here's my initial attempt at the checkstyle reporting implementation (just went off of what the twigcs package was doing).  It's currently missing the source in the XML which is optional but might be nice to know the rule that caused the violation.  Not sure how big of a lift it would be to add that to the SniffViolation.

### DIFF
--- a/src/Report/Reporter/CheckstyleReporter.php
+++ b/src/Report/Reporter/CheckstyleReporter.php
@@ -1,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
+
+namespace TwigCsFixer\Report\Reporter;
+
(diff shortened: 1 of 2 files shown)


VincentLanglet:
Hi ! Can you reopen the PR on main please ? :)

Also does https://github.com/VincentLanglet/Twig-CS-Fixer/pull/150 could help you for the sniff name ?

j-barnes:
> Hi ! Can you reopen the PR on main please ? :)
> 
> Also does #150 could help you for the sniff name ?

Will do, and #150 would be perfect and would solve that issue!  Thanks again!

PR #153, opened 2023-11-24
Add JUnit reporter

VincentLanglet:


VincentLanglet:
@j-barnes Do you also use JUnit report ? I'd like your review on this if possible
I'm not familiar with the format.

j-barnes:
@VincentLanglet  We currently do not use the JUnit Report, but as a little test I swapped out our cs2pr action for a JUnit one and it's working great!
```
  - name: Check Twig Standards
    run: |
      if [ -n "${{ steps.changed-twig-files.outputs.added_modified_renamed }}" ]; then
        vendor/bin/twig-cs-fixer lint --no-cache --report=junit ${{ steps.changed-twig-files.outputs.added_modified_renamed }} > twigcs-report.xml
      fi

  - name: Publish Test Report
    uses: mikepenz/action-junit-report@v4
    if: always() || failure()
    with:
      annotate_only: true
      report_paths: twigcs-report.xml
```

image

VincentLanglet:
Thanks for trying out.

I plan to introduce a GithubReporter like [this](https://github.com/phpstan/phpstan-src/blob/1.11.x/src/Command/ErrorFormatter/GithubErrorFormatter.php)
https://help.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-error-message

I also need to add an entry to the README. I'm open for help on this, if you want to, because I'm not the best english speaker.

Then I'll do a big new release :)

j-barnes:
That would be fantastic, and I'm always happy to help!

PR #173, opened 2024-01-27
Add rule about file double extension (.html.twig)

smnandre:
Basic attempt to enforce the "double extension" convention

> Define two extensions for filenames (e.g. index.html.twig or blog_posts.xml.twig) being the first extension (html, xml, etc.) the final format that the template will generate.

https://symfony.com/doc/current/templates.html#template-naming


docs/rules.md
```diff
@@ -36,3 +36,4 @@
 - Twig
 - DirectoryNameRule
 - FileNameRule
+- FileExtensionRule
```
VincentLanglet, line 39: You need to add the extension in src/Standard/Symfony.php.

src/Rules/File/FileExtensionRule.php
```diff
@@ -0,0 +1,73 @@
...
+ * Ensures file name uses two extensions (e.g. index.html.twig).
+ */
+final class FileExtensionRule extends AbstractRule implements ConfigurableRuleInterface
+{
+    /**
+     * @param array<string> $ignoredSubDirectories
+     */
+    public function __construct(
+        private ?string $baseDirectory = null,
+        private array $ignoredSubDirectories = [],
```
VincentLanglet, line 21: For the FileNameRule and the DirectoryNameRule, I needed:
- A base directory to start validating the path at **/templates/, and not directory before
- A ignoredSubDirectories because the `templates/bundles` doesn't follow the same convention https://symfony.com/doc/current/controller/error_pages.html

But here, I'm not sure we need the same configuration.
WDYT about starting without @smnandre ?
We can still add the config later if needed.

You'll just call
```
$fileName = FileHelper::getFileName(
    $token->getFilename(),
);
```

src/Rules/File/FileExtensionRule.php
```diff
@@ -0,0 +1,73 @@
...
+            $this->addFileError(
+                sprintf('The file must use two extensions; found ".%s".', $fileExtension),
+                $token,
+            );
+
+            return;
+        }
+
+        if (1 !== preg_match('/^[a-z][a-z0-9]{1,3}$/', $formatExtension)) {
+            $this->addFileError(
```
VincentLanglet, line 67: If you use multiple error in the same file you need to add identifier to allow ignoring one or the other.
```
$this->addFileError(
                $message,
                $token,
                'Missing'
);
```
and
```
$this->addFileError(
                $message,
                $token,
                'Invalid'
);
```
for instance.

src/Rules/File/FileExtensionRule.php
```diff
@@ -0,0 +1,73 @@
...
+        if ([] === $fileParts || null === $formatExtension) {
+            $this->addFileError(
+                sprintf('The file must use two extensions; found ".%s".', $fileExtension),
+                $token,
+            );
+
+            return;
+        }
+
+        if (1 !== preg_match('/^[a-z][a-z0-9]{1,3}$/', $formatExtension)) {
```
VincentLanglet, line 66: Do you have something about a "Valid extension" ? 
- I found nothing about `[a-z][a-z0-9]{1,3}` on https://en.wikipedia.org/wiki/Filename_extension
- I found nothing about the "extension format" on symfony standard
- I found multiple files extensions with at least 5 characters

Maybe it's better to not do this check ?

VincentLanglet (changes requested):

src/Rules/File/FileExtensionRule.php
```diff
@@ -0,0 +1,73 @@
...
+
+        $fileParts = explode('.', $fileName);
+        $fileExtension = array_pop($fileParts);
+        if ('twig' !== $fileExtension) {
+            // Not a Twig file: skip
+            return;
+        }
+
+        $formatExtension = array_pop($fileParts);
+        if ([] === $fileParts || null === $formatExtension) {
```
VincentLanglet, line 57: I'm not sure about the `[] === $fileParts`.

If understand correctly the checks, it means the file is `.foo.twig`, then `foo` is not "the second extension" but the name of the dotfile. Should we report an error for this ? What if someone use `.gitignore.twig` like https://github.com/sonata-project/dev-kit/blob/master/templates/project/.gitignore.twig but maybe they just should ignore the rule because they also have `Makefile.twig` https://github.com/sonata-project/dev-kit/blob/master/templates/project/Makefile.twig...

Let's keep this check then.
But can you add a test `.dotfile.twig` and `.dotfile.html.twig` ?

src/Rules/File/FileExtensionRule.php
smnandre: I started this PR with Mime and some checks about the content of the templates.... then realized it was waay overkill. So i tried to list all text extensions but that made no more sense. Then finally went for this... and i'm kinda relieved by your suggestion :)

smnandre: If necessary we can always add later a [a-z0-9]+ check.

src/Rules/File/FileExtensionRule.php
smnandre: Well it is not needed anymore.. but i'll keep it in mind for the next ones :)

src/Rules/File/FileExtensionRule.php
smnandre: Of course ! And thx for the explanations :)

docs/rules.md
smnandre: Do you have here some automation system / CI to check it on a Symfony app/demo ?

Or i test manually on a new webapp just to be sure Symfony... follows Symfony conventions ? 👼

VincentLanglet: I dont have any automation yet for Symfony standard no

smnandre: Good for symfony (core repo)

smnandre:
Should i add a test for the ".twig" extension or is it something assumed ?

VincentLanglet (approved):

VincentLanglet:
> Should i add a test for the ".twig" extension or is it something assumed ?

The linter only run on `.twig` file because of the finder https://github.com/VincentLanglet/Twig-CS-Fixer/blob/main/src/File/Finder.php#L17

But I like the https://github.com/VincentLanglet/Twig-CS-Fixer/pull/173/files#diff-c5c593ebdbc524bdfc583f8f378ed4059f418773fb6d6addcb72ce3a794b7980R26 check, if the file doesn't end by `.twig` the rule has no meaning.

And it's tested with `testRuleIgnoredFile`.

PR #175, opened 2024-01-27
Change behavior for dotfile in FileExtensionRule

VincentLanglet:


tests/Rules/File/FileExtension/FileExtensionRuleTest.php
```diff
@@ -21,21 +21,16 @@ public function testRuleIgnoredFile(): void
...
     {
-        $this->checkRule(new FileExtensionRule(), [], __DIR__.'/file_extension_rule_test.html.twig');
+        $this->checkRule(new FileExtensionRule(), [], __DIR__.'/FileExtensionRuleTest.html.twig');
     }
 
-    public function testRuleValidDotFile(): void
+    public function testRuleInvalidDotFile(): void
     {
-        $this->checkRule(new FileExtensionRule(), [], __DIR__.'/.dotfile.twig');
+        $this->checkRule(new FileExtensionRule(), ['FileExtension.Error'], __DIR__.'/.dotfile.twig');
```
VincentLanglet, line 29: Do you agree with this change @smnandre ?

Following the symfony standard a dot file `.dotfile.twig` doesn't have two extensions but only one so it should be considered as invalid.

smnandre: Oops sorry i may have missunderstood what you wanted :) 

Seems legit indeed

Issue #140, opened 2023-10-18
[RFC] new fixer: include function over the include tag


94noni:
Hello, a big fan of this fixer and I just saw [this PR ](https://github.com/symfony/symfony/pull/52126), what do you think of this and can it be handled via this tool?

thank you

VincentLanglet:
Hi, it is technically doable to change:
```
{% include '@WebProfiler/Profiler/table.html.twig' with { data: collector.applicationInputs, labels: ['Input', 'Value'], maxDepth: 2 } only %}
```
to
```
{{ include('@WebProfiler/Profiler/table.html.twig', { data: collector.applicationInputs, labels: ['Input', 'Value'], maxDepth: 2 }, with_context=false) }}
```

```
{% include
```
is considered as a `BLOCK_TAG_TYPE` token with the `include` value.
if you listen those token, you can
- replace it by a `{{ include(` token
- look for the `%}` token
- if there is only, replace `only %}` by `, with_context=false) }}`
- if there is not only, replace `%}` by `) }}`

Feel free to implement this custom fixer.

VincentLanglet:
Solved by https://github.com/VincentLanglet/Twig-CS-Fixer/pull/176

PR #176, opened 2024-01-27
Add Rule IncludeFunction

smnandre:
Attempt to implement #140 

Open to any feedback :)

### DIFF
--- a/src/Rules/Function/IncludeFunctionRule.php
+++ b/src/Rules/Function/IncludeFunctionRule.php
@@ -36,16 +36,15 @@ final class IncludeFunctionRule extends AbstractFixableRule
         Assert::notFalse($closingTag, 'Closing tag cannot be null.');
 
         $fixer->beginChangeSet();
-        $fixer->replaceToken($openingTag, '{{');
+
+        // Replace opening tag (and keep eventual whitespace modifiers)
+        $fixer->replaceToken($openingTag, str_replace('{%', '{{', $tokens[$openingTag]->getValue()));
         $fixer->replaceToken($tokenPosition, 'include(');
-        if ($this->isTokenMatching($tokens[$tokenPosition + 1], Token::WHITESPACE_TOKENS)) {
-            $fixer->replaceToken($tokenPosition + 1, '');
-        }
         $ignoreMissing = false;
         $withoutContext = false;
         foreach (range($tokenPosition, $closingTag - 1) as $position) {
             $token = $tokens[$position];
-            if (Token::BLOCK_NAME_TYPE === $token->getType() || Token::NAME_TYPE === $token->getType()) {
+            if (Token::NAME_TYPE === $token->getType()) {
                 switch ($token->getValue()) {
                     case 'with':
                         $fixer->replaceToken($position, ',');
@@ -66,12 +65,12 @@ final class IncludeFunctionRule extends AbstractFixableRule
                 }
             }
         }
-        $endInclude = ') }}';
-        if ($withoutContext) {
-            $endInclude = ', with_context = false'.$endInclude;
-        }
+        $endInclude = ') '. str_replace('%}', '}}', $tokens[$closingTag]->getValue());
         if ($ignoreMissing) {
             $endInclude = ', ignore_missing = true'.$endInclude;
+        }
+        if ($withoutContext) {
+            $endInclude = ', with_context = false'.$endInclude;
         }
         $fixer->replaceToken($closingTag, $endInclude);
         $fixer->endChangeSet();
--- a/src/Rules/Function/IncludeFunctionRule.php
+++ b/src/Rules/Function/IncludeFunctionRule.php
@@ -33,7 +33,7 @@ final class IncludeFunctionRule extends AbstractFixableRule
         Assert::notFalse($openingTag, 'Opening tag cannot be null.');
 
         $closingTag = $this->findNext(Token::BLOCK_END_TYPE, $tokens, $tokenPosition);
-        Assert::notFalse($closingTag, 'Closing tag cannot not be null.');
+        Assert::notFalse($closingTag, 'Closing tag cannot be null.');
 
         $fixer->beginChangeSet();
         $fixer->replaceToken($openingTag, '{{');
--- a/src/Rules/Function/IncludeFunctionRule.php
+++ b/src/Rules/Function/IncludeFunctionRule.php
@@ -1,0 +1,79 @@
+<?php
+
+declare(strict_types=1);
+
+namespace TwigCsFixer\Rules\Function;
+
+use TwigCsFixer\Rules\AbstractFixableRule;
+use TwigCsFixer\Token\Token;
+use Webmozart\Assert\Assert;
+
+/**
+ * Ensures that include function is used instead of function tag.
+ */
+final class IncludeFunctionRule extends AbstractFixableRule
+{
+    protected function process(int $tokenPosition, array $tokens): void
+    {
+        $token = $tokens[$tokenPosition];
+        if (!$this->isTokenMatching($token, Token::BLOCK_NAME_TYPE, 'include')) {
+            return;
+        }
+
+        $openingTag = $this->findPrevious(Token::BLOCK_START_TYPE, $tokens, $tokenPosition);
+        Assert::notFalse($openingTag, 'Opening tag cannot not be null.');
+
+        $closingTag = $this->findNext(Token::BLOCK_END_TYPE, $tokens, $tokenPosition);
+        Assert::notFalse($closingTag, 'Closing tag cannot not be null.');
+
+        $fixer = $this->addFixableError(
+            'Include function must be used instead of include tag.',
+            $token
+        );
+
+        if (null === $fixer) {
+            return;
+        }
+
+        $fixer->beginChangeSet();
+        $fixer->replaceToken($openingTag, '{{');
+        $fixer->replaceToken($tokenPosition, 'include(');
+        if ($this->isTokenMatching($tokens[$tokenPosition + 1], Token::WHITESPACE_TOKENS)) {
+            $fixer->replaceToken($tokenPosition + 1, '');
+        }
+        $ignoreMissing = false;
+        $withoutContext = false;
+        foreach (range($tokenPosition, $closingTag - 1) as $position) {
+            $token = $tokens[$position];
+            if (Token::BLOCK_NAME_TYPE === $token->getType() || Token::NAME_TYPE === $token->getType()) {
+                switch ($token->getValue()) {
+                    case 'with':
+                        $fixer->replaceToken($position, ',');
+                        break;
+                    case 'only':
+                        $withoutContext = true;
+                        $fixer->replaceToken($position, '');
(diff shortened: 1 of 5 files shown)


src/Rules/Function/IncludeFunctionRule.php
```diff
@@ -0,0 +1,79 @@
...
+{
+    protected function process(int $tokenPosition, array $tokens): void
+    {
+        $token = $tokens[$tokenPosition];
+        if (!$this->isTokenMatching($token, Token::BLOCK_NAME_TYPE, 'include')) {
+            return;
+        }
+
+        $openingTag = $this->findPrevious(Token::BLOCK_START_TYPE, $tokens, $tokenPosition);
+        Assert::notFalse($openingTag, 'Opening tag cannot not be null.');
```
VincentLanglet, line 24:
```suggestion
        Assert::notFalse($openingTag, 'Opening tag cannot be null.');
```

src/Rules/Function/IncludeFunctionRule.php
```diff
@@ -0,0 +1,79 @@
...
+
+        $closingTag = $this->findNext(Token::BLOCK_END_TYPE, $tokens, $tokenPosition);
+        Assert::notFalse($closingTag, 'Closing tag cannot not be null.');
```
VincentLanglet, line 27:
```suggestion
        Assert::notFalse($closingTag, 'Closing tag cannot be null.');
```

src/Rules/Function/IncludeFunctionRule.php
```diff
@@ -0,0 +1,79 @@
...
+final class IncludeFunctionRule extends AbstractFixableRule
+{
+    protected function process(int $tokenPosition, array $tokens): void
+    {
+        $token = $tokens[$tokenPosition];
+        if (!$this->isTokenMatching($token, Token::BLOCK_NAME_TYPE, 'include')) {
+            return;
+        }
+
+        $openingTag = $this->findPrevious(Token::BLOCK_START_TYPE, $tokens, $tokenPosition);
```
VincentLanglet, line 23: You can look for the opening tag and ending tag after the
```
if (null === $fixer) {
            return;
        }
```

check.

src/Rules/Function/IncludeFunctionRule.php
```diff
@@ -0,0 +1,79 @@
...
+            'Include function must be used instead of include tag.',
+            $token
+        );
+
+        if (null === $fixer) {
+            return;
+        }
+
+        $fixer->beginChangeSet();
+        $fixer->replaceToken($openingTag, '{{');
```
VincentLanglet, line 39: You're changing `{%` to `{{`.
What if it was `{%~` or `{%-` ? (This will need to be tested)

src/Rules/Function/IncludeFunctionRule.php
```diff
@@ -0,0 +1,79 @@
...
+                        break;
+                    case 'missing':
+                        $fixer->replaceToken($position, '');
+                        break;
+                    default:
+                        break;
+                }
+            }
+        }
+        $endInclude = ') }}';
```
VincentLanglet, line 69: Same, we will need to support `~%}` and `-%}`

src/Rules/Function/IncludeFunctionRule.php
```diff
@@ -0,0 +1,79 @@
...
+        if ($withoutContext) {
+            $endInclude = ', with_context = false'.$endInclude;
+        }
+        if ($ignoreMissing) {
+            $endInclude = ', ignore_missing = true'.$endInclude;
+        }
```
VincentLanglet, line 75: I'm not sure about named parameters. I didn't see them often used in twig.

So I wonder if we shouldn't try to have
```
include('template.html' , vars  , false)
```
rather than
```
include('template.html' , vars  , with_context = false)
```
and same for others options.

tests/Rules/Function/IncludeFunction/IncludeFunctionRuleTest.fixed.twig
```diff
@@ -0,0 +1,14 @@
+{{ include('template.html' ) }}
+{{ include(['template_a.html', 'template_b.html']   , {'foo': 'bar'}  , ignore_missing = true, with_context = false) }}
+{{ include('template.html'  , with_context = false) }}
+{{ include('template.html' , vars ) }}
+{{ include('template.html' , vars  , with_context = false) }}
+{{ include('template.html' , {'foo': 'bar'} ) }}
+{{ include('template.html' , {'foo': 'bar'}  , with_context = false) }}
```
VincentLanglet, line 7: You started to handled unnecessary whitespace with your 
```
if ($this->isTokenMatching($tokens[$tokenPosition + 1], Token::WHITESPACE_TOKENS)) {
            $fixer->replaceToken($tokenPosition + 1, '');
        }
```
check, but we still have unnecessary one here, before the `,`.

Imho, we should either
- complicate the rule to handle correctly ALL the whitespace
- or handle none

src/Rules/Function/IncludeFunctionRule.php
```diff
@@ -0,0 +1,79 @@
...
+        $fixer->replaceToken($openingTag, '{{');
+        $fixer->replaceToken($tokenPosition, 'include(');
+        if ($this->isTokenMatching($tokens[$tokenPosition + 1], Token::WHITESPACE_TOKENS)) {
+            $fixer->replaceToken($tokenPosition + 1, '');
+        }
+        $ignoreMissing = false;
+        $withoutContext = false;
+        foreach (range($tokenPosition, $closingTag - 1) as $position) {
+            $token = $tokens[$position];
+            if (Token::BLOCK_NAME_TYPE === $token->getType() || Token::NAME_TYPE === $token->getType()) {
```
VincentLanglet, line 48: Why do you need both the block name type and the name type ?
Can 4 of this keyword be found in the 2 differents types (or should we split the switch/case) ?

src/Rules/Function/IncludeFunctionRule.php
smnandre: Thank you

src/Rules/Function/IncludeFunctionRule.php
smnandre: Thx i keep forgetting about those

tests/Rules/Function/IncludeFunction/IncludeFunctionRuleTest.fixed.twig
smnandre: I'd say none, even if it's a bit weird-looking in the "fixed.twig" files... (i went back and forth on this when i coded)

Do you have any preference ?

src/Rules/Function/IncludeFunctionRule.php
smnandre: To be honest I have absolutely no idea wich of those two args (context / missing) is the first one / second one .. 

Tried to follow "blindly" the doc : https://twig.symfony.com/doc/3.x/functions/include.html 

I remove them and it can be the goal of another rule i guess later

smnandre: Oh well..   we  generate "`include(...,  false, true)`" if there was only "`ignore_missing`" ? :| 

```php
function twig_include(Environment $env, $context, $template, $variables = [], $withContext = true, $ignoreMissing = false, $sandboxed = false)
{
```

src/Rules/Function/IncludeFunctionRule.php
smnandre: First TYPE was useless. But we need to check both name and type to avoid false positive with vars or constants :/

tests/Rules/Function/IncludeFunction/IncludeFunctionRuleTest.fixed.twig
smnandre: (removed from now, tell me if you want otherwise)

src/Rules/Function/IncludeFunctionRule.php
VincentLanglet: Yes, I had in mind to generate

Include ($template, [], false, true)

If the only option passed was "ignore missing".

I agree that "false" is unnecessary passed but I feel like people are not used to named parameter yet.

tests/Rules/Function/IncludeFunction/IncludeFunctionRuleTest.fixed.twig
VincentLanglet: None seems the simpler.

We can still iterate on it later if someone complains about it.

Our recommendation would to use both the include fixer with all the other spacing fixer

src/Rules/Function/IncludeFunctionRule.php
smnandre: Done !

VincentLanglet (approved):

src/Rules/Function/IncludeFunctionRule.php
```diff
@@ -0,0 +1,78 @@
...
+                        $fixer->replaceToken($position, '');
+                        break;
+                    default:
+                        break;
+                }
+            }
+        }
+        $endInclude = ') '.str_replace('%}', '}}', $tokens[$closingTag]->getValue());
+        if ($ignoreMissing) {
+            $endInclude = sprintf(', %s', $ignoreMissing ? 'true' : 'false').$endInclude;
```
VincentLanglet, line 70: The ternary condition here is always true

VincentLanglet (changes requested):

src/Rules/Function/IncludeFunctionRule.php
smnandre: 😶‍🌫️

smnandre: Fixed, sorry

VincentLanglet (approved):

VincentLanglet:
Thanks !

tests/Rules/Function/IncludeFunction/IncludeFunctionRuleTest.twig
```diff
@@ -0,0 +1,16 @@
+{% include 'template.html' %}
+{% include ['template_a.html', 'template_b.html'] ignore missing with {'foo': 'bar'} only %}
+{% include 'template.html' only %}
```
VincentLanglet, line 3: @smnandre This is fixed in `{{ include( 'template.html'  , false) }}`.

I think it's wrong, and should be `{{ include( 'template.html', [], false) }}`
Not sure if it's easy without named parameters...

VincentLanglet: Solved in https://github.com/VincentLanglet/Twig-CS-Fixer/pull/178

94noni:
Waw nice one @smnandre and @VincentLanglet 👌🏼

tests/Rules/Function/IncludeFunction/IncludeFunctionRuleTest.twig
smnandre: Damn really sorry to let so many mistakes :(

PR #177, opened 2024-01-28
Improve include rule

VincentLanglet:
I was curious about how hard it would be to handle whitespace in the include rules, and that doesn't seems that hard

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -19,9 +19,9 @@
 ### Non-fixable
 
 - **DirectoryNameRule**: ensures that directory name uses snake_case (configurable).
+- **FileExtensionRule**: ensures that file name uses two extensions (e.g. index.html.twig).
 - **FileNameRule**: ensures that file name uses snake_case (configurable).
 - **VariableNameRule**: ensures that variable name uses snake_case (configurable).
-- **FileExtensionRule**: ensures that file name uses two extensions (e.g. index.html.twig).
 
 ## Standards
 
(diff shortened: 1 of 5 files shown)


VincentLanglet:
@smnandre I tried to play with the rule to see if it was easy to handle whitespace. Do you think it's worth it to add this ?

smnandre:
To be honest... i don't know (not a false-way of saying i don't like... i really don"t know)

Option A: a fixer does only one thing
- pros: when "atomic", it's easier to test, maintain, update, etc... 
- cons: code will really soon become duplicated in a lot of places, and become less readable

Option B: a fixer has to produce correct code
- pros: easier to test, no nested calls, etc. 
- cons: tests are ugly and force us to expect weird things like multiple spaces

Maybe there is a "C" option which would be a mix of A and having test helpers that do call others fixers before comparing expected / produced ?

VincentLanglet:
> To be honest... i don't know (not a false-way of saying i don't like... i really don"t know)
> 
> Option A: a fixer does only one thing
> 
> * pros: when "atomic", it's easier to test, maintain, update, etc...
> * cons: code will really soon become duplicated in a lot of places, and become less readable
> 
> Option B: a fixer has to produce correct code
> 
> * pros: easier to test, no nested calls, etc.
> * cons: tests are ugly and force us to expect weird things like multiple spaces
> 
> Maybe there is a "C" option which would be a mix of A and having test helpers that do call others fixers before comparing expected / produced ?

The C option is not a bad idea: this would be 
https://github.com/VincentLanglet/Twig-CS-Fixer/pull/179/files if I understand correctly the suggestion

smnandre:
Oh this is cleaner and simpler than what i had in mind 👏

PR #218, opened 2024-04-30
feat: add `IncludeFunctionRule` in default standard fixer

94noni:
hey, 
I know there is https://github.com/VincentLanglet/Twig-CS-Fixer/blob/main/docs/configuration.md
but I see a lot of codebase going to this, which is more correct in twig 

wdyt of it being standard?

### DIFF
--- a/src/Standard/TwigCsFixer.php
+++ b/src/Standard/TwigCsFixer.php
@@ -25,7 +25,7 @@ final class TwigCsFixer implements StandardInterface
             new BlankEOFRule(),
             new BlockNameSpacingRule(),
             new EmptyLinesRule(),
-            new IncludeFunctionRule();
+            new IncludeFunctionRule(),
             new IndentRule(),
             new SingleQuoteRule(),
             new TrailingCommaSingleLineRule(),
--- a/src/Standard/TwigCsFixer.php
+++ b/src/Standard/TwigCsFixer.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace TwigCsFixer\Standard;
 
 use TwigCsFixer\Rules\Delimiter\BlockNameSpacingRule;
+use TwigCsFixer\Rules\Function\IncludeFunctionRule;
 use TwigCsFixer\Rules\Punctuation\TrailingCommaSingleLineRule;
 use TwigCsFixer\Rules\String\SingleQuoteRule;
 use TwigCsFixer\Rules\Whitespace\BlankEOFRule;
@@ -24,6 +25,7 @@ final class TwigCsFixer implements StandardInterface
             new BlankEOFRule(),
             new BlockNameSpacingRule(),
             new EmptyLinesRule(),
+            new IncludeFunctionRule();
             new IndentRule(),
             new SingleQuoteRule(),
             new TrailingCommaSingleLineRule(),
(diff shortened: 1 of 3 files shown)


src/Standard/TwigCsFixer.php
```diff
@@ -28,6 +29,7 @@ public function getRules(): array
             new SingleQuoteRule(),
             new TrailingCommaSingleLineRule(),
             new TrailingSpaceRule(),
+            new IncludeFunctionRule(),
```
VincentLanglet, line 32: I try to order rules alphabetically, can you do it too ?

Also, you will need to update the tests.

VincentLanglet (changes requested):

94noni:
> Can you list some example of codebase ?

ref https://twitter.com/lyrixx/status/1765694838756807153
ref https://github.com/VincentLanglet/Twig-CS-Fixer/issues/140

for the ones i had in my mind 

---

could event be usefull to propose to add such tool in the symfony/symfony codebase, as there is the php cs fixer already ^^

src/Standard/TwigCsFixer.php
94noni: I will do for sure, I wanted to know if its ok for you first

src/Standard/TwigCsFixer.php
```diff
@@ -24,6 +25,7 @@ public function getRules(): array
             new BlankEOFRule(),
             new BlockNameSpacingRule(),
             new EmptyLinesRule(),
+            new IncludeFunctionRule();
```
VincentLanglet, line 28:
```suggestion
            new IncludeFunctionRule(),
```

94noni: Oops mybad

VincentLanglet (approved):
I'll think a little before merging this.

I dunno if I want to provide this by default.
Until the tag is deprecated, this might be considered as very opinionated...

Also, I could introduce another standard for such rules...

94noni:
> Until the tag is deprecated

do you think its worth it an issue is opened on the symfony/twig repos for the deprecation?

> this might be considered as very opinionated...

indeed it can be seen like this, but its the real usage in fine
also the official twig doc `note` this usage as a `semantically more "correct"`
https://twig.symfony.com/doc/3.x/tags/include.html

VincentLanglet:
> > Until the tag is deprecated
> 
> do you think its worth it an issue is opened on the symfony/twig repos for the deprecation?

Sure, see https://twitter.com/fabpot/status/1765719169662648337

94noni:
> Sure, see https://twitter.com/fabpot/status/1765719169662648337

I've read this of course, thus this PR
my point was, do you think we (me/you) should create and issue or propose a PR directly on symfony/twig repos ?

does doing this will make this PR advance and your comment `this might be considered as very opinionated...` more real ?

thank you

VincentLanglet:
> I've read this of course, thus this PR my point was, do you think we (me/you) should create and issue or propose a PR directly on symfony/twig repos ?

Sure, I think it's a good idea to propose a PR on twig.

> does doing this will make this PR advance and your comment `this might be considered as very opinionated...` more real ?

If the deprecation is accepted on twig repository, I'll merge 100% this PR.
But it's not a blocker, I think I'll merge it anyway.

94noni:
I've just opened an issue https://github.com/twigphp/Twig/issues/4067
just to know the process of code deprecating of twig repo (I am not used to it)

lets see how it goes 👍🏻

PR #240, opened 2024-06-19
Introduce alias for command

### LINKED ISSUE #223: [DX] Aliases "check" & "fix"
i'd like to have two aliases: `check` & `fix `, so i started to see how that could be done.

This seems to work perfectly by adding thoses aliases in the entry point, and two lines in the command 

```php
// bin/twig-cs-fixer

// ...

$command = new TwigCsFixerCommand();
$command->setAliases(['fix', 'check']);

$application = new Application();
```

```php
// TwigCsFixerCommand.php


    protected function execute(InputInterface $input, OutputInterface $output): int
    {
        if ('check' === $input->getArgument('command')) {
            $input->setOption('fix', false);
        } elseif ('fix' === $input->getArgument('command')) {
            $input->setOption('fix', true);
        }
```

Would you be open to that ? If yes do you think there is a better way to implement it ?

VincentLanglet:
Closes #223


smnandre:
Tested right now, works perfectly... and thank you!

VincentLanglet:
> Tested right now, works perfectly... and thank you!

Thanks for the feedback

Issue #237, opened 2024-06-04
GitHub reporter should always return relative filenames


ruudk:
When using a custom Finder that scan 2 specific directies, the filenames reported become absolute instead of relative.

```php
$finder = TwigCsFixer\File\Finder::create()
    ->in(__DIR__ . '/templates')
    ->in(__DIR__ . '/src/Core/Operations/Admin/Resources/views');

$config = new TwigCsFixer\Config\Config();
$config->setFinder($finder);

return $config;
```

```command
$ vendor/bin/twig-cs-fixer lint --report=github

::error file=/Volumes/CS/www/symfony/templates/mail/welcome.html.twig,line=106,col=135::Expecting 1 whitespace before "}"; found 0
```

In order for GitHub to process these properly, they should be relative to the repository root.

Any idea how to deal with that?

If I remove the custom `in()` directories, it works again.

VincentLanglet:
Hi, 

The filePath is coming from this line 
https://github.com/VincentLanglet/Twig-CS-Fixer/blob/30fef7f8dc01bf1eaf948400a086ee2becae0f15/src/Runner/Linter.php#L44
and then, it's passed without any modification
- to the tokens
- to the violations
- to the report

> If I remove the custom in() directories, it works again.

The fact it works without the custom in, would say it's more related to the Symfony FInder than "my" lib.

In you code example, you're using non-relative path
```
$finder = TwigCsFixer\File\Finder::create()
    ->in(__DIR__ . '/templates')
    ->in(__DIR__ . '/src/Core/Operations/Admin/Resources/views');
```
I assume Symfony copy this. It can be solve by using the config
```
$finder = TwigCsFixer\File\Finder::create()
    ->in('templates')
    ->in('src/Core/Operations/Admin/Resources/views');
```

I'm not sure about how works the Finder, but the best would be a method on the Finder to select if we want relative or absolute path. WDYT ?

VincentLanglet:
I consider this solved by
```
$finder = TwigCsFixer\File\Finder::create()
    ->in('templates')
    ->in('src/Core/Operations/Admin/Resources/views');
```

PR #241, opened 2024-06-25
Add example code to demonstrate how to set templates-folder

k00ni:
The documentation is missing some example code which demonstrates how to set the templates folder (**Files** section). In a common PHP project I usually have my Twig templates located in a sub-folder, e.g. `templates` and I don't want Twig-CS-Fixer to lint all the files in the current directory, as [this part suggests](https://github.com/VincentLanglet/Twig-CS-Fixer/blob/main/docs/configuration.md#files).

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -73,7 +73,7 @@ If you want a more sophisticated rule, you can configure it in the config file:
 <?php
 
 $finder = new TwigCsFixer\File\Finder();
-$finder->in(__DIR__.'/templates'); // only lint files in the given directory
+$finder->in('templates');
 $finder->exclude('myCustomDirectory');
 
 $config = new TwigCsFixer\Config\Config();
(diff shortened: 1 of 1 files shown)


docs/configuration.md
```diff
@@ -73,6 +73,7 @@ If you want a more sophisticated rule, you can configure it in the config file:
 <?php
 
 $finder = new TwigCsFixer\File\Finder();
+$finder->in(__DIR__.'/templates'); // only lint files in the given directory
```
VincentLanglet, line 76:
```suggestion
$finder->in('templates');
```
Relatives files should be preferred to avoid issues like https://github.com/VincentLanglet/Twig-CS-Fixer/issues/237

Also, I'm not sure the comment is needed, the method call seems self-explanatory

VincentLanglet (changes requested):

docs/configuration.md
k00ni: Well, I didn't know about Symfony Finder so I had to search a bit before I was able to set it up. The [in function](https://symfony.com/doc/current/components/finder.html#location) isn't that self explanatory in my opinion, that's why I added the comment.

Anyway, feel free to change whatever you want.

VincentLanglet (approved):

VincentLanglet:
Thanks

PR #247, opened 2024-07-03
Remove excess spacing before multiline block

### LINKED ISSUE #245: Remove excess spacing before multiline block `%}`
### Rule(s) related to or rule(s) missing

Using the standard TwigCsFixer rules I notice something odd.

### Actual behavior

```twig
{% trans 'my.key' with {
    faqUrl: 'http://faq',
    contactUrl: 'http://contact',
    }

%}
```

### Expected behavior

```twig
{% trans 'my.key' with {
    faqUrl: 'http://faq',
    contactUrl: 'http://contact',
    } %}
```

Why isn't the excess newline and spacing not removed?

ruudk:
Fixes #245


VincentLanglet:
If you want to take a look, I added a commit which seems to fix everything except
```
Error: Cannot unpack array with string keys
```
which is a PHP 8 specific error.

Some shared thought:
I'm not sure the AbstractSpacingRule should `implements ConfigurableRuleInterface` by default. We might prefer to enable the config manually on the different rules. This PR is adding `skipIfNewLine` option for the rules:
  - OperatorSpacingRule
  - DelimiterSpacingRule
  - BlockNameSpacingRule
  - PunctuationSpacingRule

Maybe it's better to just make the DelimiterSpacingRule configurable for now, and add
```
public function __construct() {
    parent::__construct();
}
```
to other rules in order to remove the param from them (so far) and wait for feature request (?) This way we could know
- if the option is required
- if the option should be the first param of the rule (as an example PunctuationSpacingRule already has option).

Does it seem ok to you ?

ruudk:
@VincentLanglet Thanks so much for helping me out! I agree with what you're saying. I just rebased the PR and made the suggested changes. Let me know what you think!

src/Rules/AbstractSpacingRule.php
```diff
@@ -11,6 +11,17 @@
...
+    public function __construct(private bool $skipIfNewLine = true)
+    {
+    }
+
+    public function getConfiguration(): array
+    {
+        return [
+            'skipIfNewLine' => $this->skipIfNewLine,
+        ];
+    }
```
VincentLanglet, line 23: We can simplify to

```suggestion
    protected bool $skipIfNewLine = false;
```
And setting the value in the construct of DelimiterSpacingRule with the implementation of getConfiguration

src/Rules/Punctuation/PunctuationSpacingRule.php
```diff
@@ -47,13 +47,15 @@ public function __construct(
...
+        parent::__construct();
+
         $this->punctuationWithSpaceBefore = $punctuationWithSpaceBefore + self::DEFAULT_SPACE_BEFORE;
         $this->punctuationWithSpaceAfter = $punctuationWithSpaceAfter + self::DEFAULT_SPACE_AFTER;
     }
 
     public function getConfiguration(): array
     {
-        return [
+        return parent::getConfiguration() + [
```
VincentLanglet, line 58:
```suggestion
        return [
```

src/Rules/Punctuation/PunctuationSpacingRule.php
```diff
@@ -47,13 +47,15 @@ public function __construct(
         array $punctuationWithSpaceBefore = [],
         array $punctuationWithSpaceAfter = [],
     ) {
+        parent::__construct();
```
VincentLanglet, line 50: This won't be needed with the previous suggestion

tests/Rules/Punctuation/PunctuationSpacing/PunctuationSpacingRuleTest.php
```diff
@@ -13,6 +13,7 @@ public function testConfiguration(): void
     {
         static::assertSame(
             [
+                'skipIfNewLine' => true,
```
VincentLanglet, line 16: Shouldn't be added to the config since it's unused

VincentLanglet (changes requested):

ruudk:
Applied the feedback!

ruudk:
Pushed an updated test that now fails.. I think it should also remove the spaces.

VincentLanglet:
I wont be able to debug this before the next week since I dont have a computer ATM.

VincentLanglet:
> Pushed an updated test that now fails.. I think it should also remove the spaces.

Hi @ruudk, I don't have any real failure with
```
{%

    trans with {
    contactUrl: 'http://contact',
}

    %}my.key{% endtrans %}
````
I just need to update tests/failure message.

But I think the failed test you wanted is the one with only one space
```
{%

    trans with {
    contactUrl: 'http://contact',
}

 %}my.key{% endtrans %}
```

Is it the issue you encounter ? If yes, I'll take a look.

**Comment by ruudk** [comment]
Yes, I couldn't provide a good failing test.

So I think what we want is that with `skipIfNewLine=true` (default) the following will be changed:
```diff
 {%
 
     trans with {
     contactUrl: 'http://contact',
 }
 
-    %}my.key{% endtrans %}
+ %}my.key{% endtrans %}
```
So it changes 4 spaces after to 1 space after.

And then with `skipIfNewLine=false` it should do the following:
```diff
 {%
 
     trans with {
     contactUrl: 'http://contact',
- }
- 
-    %}my.key{% endtrans %}
+ } %}my.key{% endtrans %}
```

It should take the 4 spaces + the 2 newlines, and convert that to 1 space.

ruudk:
I think `\TwigCsFixer\Tests\Rules\AbstractRuleTestCase::checkRule` can be improved.

It should not immediately fail the test when one of the diffs fails. It should print the diff, then continue with showing the diff for the violations, and at the end fail it. This way, when you are trying to create a failing test, you can see the violations too. Now, I had to comment that part out, in order to see what the violations would be.

VincentLanglet:
> So I think what we want is that with `skipIfNewLine=true` (default) the following will be changed:
> 
> ```diff
>  {%
>  
>      trans with {
>      contactUrl: 'http://contact',
>  }
>  
> -    %}my.key{% endtrans %}
> + %}my.key{% endtrans %}
> ```

Not really, with skipIfNewLine let untouched this code because there is a new line before the last non empty token and `%}`. This allow to
- Add a new line
- Indent your code after the new line

> And then with `skipIfNewLine=false` it should do the following:
> 
> ```diff
>  {%
>  
>      trans with {
>      contactUrl: 'http://contact',
> - }
> - 
> -    %}my.key{% endtrans %}
> + } %}my.key{% endtrans %}
> ```
> 
> It should take the 4 spaces + the 2 newlines, and convert that to 1 space.

It's understandable. That also mean that if you have only one space, new lines will be removed

```diff
 {%
 
     trans with {
     contactUrl: 'http://contact',
- }
- 
-  %}my.key{% endtrans %}
+ } %}my.key{% endtrans %}
```

ruudk:
> It's understandable. That also mean that if you have only one space, new lines will be removed

I think this is what I want with this feature. See newlines also as whitespace and enforce it regardless.

I never write 

```twig
{%
  if var == true 
%}
```

or

```twig
{% 
trans into 'nl'
%}my.key{%endtrans %}
```

Am I missing something? Is this common? 

Also, this rule is not default enabled, so we can make this different?

VincentLanglet:
It should be ok with the latest commit https://github.com/VincentLanglet/Twig-CS-Fixer/pull/247/commits/33fb7bb2922a61a36310ad7e28592a20409c3e6d @ruudk

ruudk:
@VincentLanglet Amazing 🎉 Thank you!

VincentLanglet (approved):

VincentLanglet:
Thanks

Issue #248, opened 2024-07-07
Verify that constants exist


ruudk:
It's possible to use the `constant('App\\SomeClass::SomeConstant')` function in to reference a constant. 

But it will only be verified on runtime.

It would be great if we could define a non-fixable rule that verifies these when doing a lint.

Now I'm not sure if this is something that should be part of Twig CS Fixer, but if it is, what would be the best way to do this? I feel like creating a NodeVisitor that finds these `constant` functions and then reports errors when the constant does not exist would be best. Otherwise I have to check this based on the tokens.

VincentLanglet:
I agree I'm not sure this lib is the best one for such rule.
It's more a "static-analysis" rule rather than a "linter" one.

ruudk:
This can now be implemented using https://github.com/VincentLanglet/Twig-CS-Fixer/issues/249 🥳

VincentLanglet:
Yes, thanks for your work.

I'll try a small refacto first https://github.com/VincentLanglet/Twig-CS-Fixer/pull/251 (WIP)

ruudk:
Yes, I have it ready on our project. Will contribute it back next week.

PR #249, opened 2024-07-09
Introduce NodeRuleInterface

### LINKED ISSUE #246: Make it possible to turn off `OptionalDependencies` in `StubbedEnvironment`
First of all, thanks for this amazing tool! 👏 

In our Symfony project we use a custom translation mechanism. We don't want people to use the default `trans` blocks and functions.

But these are automatically added because the classes exist:

https://github.com/VincentLanglet/Twig-CS-Fixer/blob/1f1bcc5b7a608c64603e8d3a879508316bfc9919/src/Environment/StubbedEnvironment.php#L128-L133

It would be great if I could turn off this behavior. What would be the best way?

Maybe introduce an option on the `Config` that's enabled by default, that we can disable?

Then I can manually add all the required token parsers myself.

I could create the PR if you tell me how you want it to be done 😊

ruudk:
Fixes #246, #248 

### Introduce NodeRuleInterface

With this you can add violations based on AST nodes.

### Introduce ForbiddenFilterRule, ForbiddenBlockRule & ForbiddenFunctionRule

This can be used to disable certain filters/blocks/functions that are available during compile/runtime.

These are just some examples. People can create other NodeVisitors for other type of verifications.

For example, we have a custom `enum` function. The value passed is a FQCN. This should be validated
early on, instead of runtime. With a NodeVisitor this could be done easily during linting.

Another idea could be to verify FQCN's for `constant` function.

Or verify that a translation key exists. Or that parameters to the translation are missing.

### DIFF
--- a/src/Runner/Linter.php
+++ b/src/Runner/Linter.php
@@ -6,6 +6,7 @@ namespace TwigCsFixer\Runner;
 
 use Twig\Environment;
 use Twig\Error\Error;
+use Twig\NodeTraverser;
 use Twig\Source;
 use TwigCsFixer\Cache\Manager\CacheManagerInterface;
 use TwigCsFixer\Cache\Manager\NullCacheManager;
@@ -14,6 +15,8 @@ use TwigCsFixer\Exception\CannotTokenizeException;
 use TwigCsFixer\Report\Report;
 use TwigCsFixer\Report\Violation;
 use TwigCsFixer\Report\ViolationId;
+use TwigCsFixer\Rules\Node\NodeRuleInterface;
+use TwigCsFixer\Rules\RuleInterface;
 use TwigCsFixer\Ruleset\Ruleset;
 use TwigCsFixer\Token\TokenizerInterface;
 
@@ -39,6 +42,18 @@ final class Linter
     {
         $report = new Report($files);
 
+        /**
+         * @var list<RuleInterface> $rules
+         */
+        $rules = array_filter($ruleset->getRules(), fn ($rule) => $rule instanceof RuleInterface);
+
+        /**
+         * @var list<NodeRuleInterface> $nodeVisitorRules
+         */
+        $nodeVisitorRules = array_filter($ruleset->getRules(), fn ($rule) => $rule instanceof NodeRuleInterface);
+
+        $traverser = new NodeTraverser($this->env, $nodeVisitorRules);
+
         // Process
         foreach ($files as $file) {
             $filePath = $file->getPathname();
@@ -60,9 +75,10 @@ final class Linter
             }
 
             // Validate the file with twig parser.
+            $node = null;
             try {
                 $twigSource = new Source($content, $filePath);
-                $this->env->parse($this->env->tokenize($twigSource));
+                $node = $this->env->parse($this->env->tokenize($twigSource));
             } catch (Error $error) {
                 $violation = new Violation(
                     Violation::LEVEL_FATAL,
@@ -81,6 +97,7 @@ final class Linter
                     $newContent = $fixer->fixFile($content, $ruleset);
                     // Don't write the file if it is unchanged in order not to invalidate mtime based caches.
                     if ($newContent !== $content) {
+                        $node = null;
                         file_put_contents($filePath, $newContent);
                         $content = $newContent;
                         $report->addFixedFile($filePath);
@@ -122,9 +139,34 @@ final class Linter
             }
             restore_error_handler();
 
-            $rules = $ruleset->getRules();
             foreach ($rules as $rule) {
                 $rule->lintFile($stream, $report, $ignoredViolations);
+            }
+
+            if ([] !== $nodeVisitorRules) {
+                if (null === $node) {
+                    try {
+                        $twigSource = new Source($content, $filePath);
+                        $node = $this->env->parse($this->env->tokenize($twigSource));
+                    } catch (Error $error) {
+                        $violation = new Violation(
+                            Violation::LEVEL_FATAL,
+                            sprintf('File is invalid: %s', $error->getRawMessage()),
+                            $filePath,
+                            null,
+                            new ViolationId(line: $error->getTemplateLine())
+                        );
+
+                        $report->addViolation($violation);
+                        continue;
+                    }
+                }
+
+                foreach ($nodeVisitorRules as $nodeVisitor) {
+                    $nodeVisitor->enterFile($report, $filePath, $ignoredViolations);
+                }
+
+                $traverser->traverse($node);
             }
 
             // Only cache the file if there is no error in order to
--- a/src/Cache/Signature.php
+++ b/src/Cache/Signature.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace TwigCsFixer\Cache;
 
 use TwigCsFixer\Rules\ConfigurableRuleInterface;
+use TwigCsFixer\Rules\Node\ConfigurableNodeRuleInterface;
 use TwigCsFixer\Ruleset\Ruleset;
 
 final class Signature
@@ -26,7 +27,7 @@ final class Signature
     ): self {
         $rules = [];
         foreach ($ruleset->getRules() as $rule) {
-            $rules[$rule::class] = $rule instanceof ConfigurableRuleInterface
+            $rules[$rule::class] = $rule instanceof ConfigurableRuleInterface || $rule instanceof ConfigurableNodeRuleInterface
                 ? $rule->getConfiguration()
                 : null;
         }
--- a/src/Rules/Node/AbstractNodeRule.php
+++ b/src/Rules/Node/AbstractNodeRule.php
@@ -1,0 +1,136 @@
+<?php
+
+declare(strict_types=1);
+
+namespace TwigCsFixer\Rules\Node;
+
+use Twig\Environment;
+use Twig\Node\Node;
+use TwigCsFixer\Report\Report;
+use TwigCsFixer\Report\Violation;
+use TwigCsFixer\Report\ViolationId;
+
+abstract class AbstractNodeRule implements NodeRuleInterface
+{
+    private ?Report $report = null;
+
+    private ?string $filePath = null;
+
+    /**
+     * @var list<ViolationId>
+     */
+    private array $ignoredViolations = [];
+
+    public function getName(): string
+    {
+        return static::class;
+    }
+
+    public function getShortName(): string
+    {
+        $shortName = (new \ReflectionClass($this))->getShortName();
+
+        return str_ends_with($shortName, 'Rule') ? substr($shortName, 0, -4) : $shortName;
+    }
+
+    public function enterFile(Report $report, string $filePath, array $ignoredViolations = []): void
+    {
+        $this->report = $report;
+        $this->filePath = $filePath;
+        $this->ignoredViolations = $ignoredViolations;
+    }
+
+    public function enterNode(Node $node, Environment $env): Node
+    {
+        return $node;
+    }
+
+    public function leaveNode(Node $node, Environment $env): Node
+    {
+        return $node;
+    }
+
+    public function getPriority(): int
+    {
+        return 0;
+    }
+
+    protected function addWarning(string $message, Node $node, ?string $messageId = null): bool
+    {
+        return $this->addMessage(
+            Violation::LEVEL_WARNING,
+            $message,
+            $node->getTemplateName(),
+            $node->getTemplateLine(),
+            null,
+            $messageId,
+        );
+    }
+
+    protected function addFileWarning(string $message, Node $node, ?string $messageId = null): bool
+    {
+        return $this->addMessage(
+            Violation::LEVEL_WARNING,
+            $message,
+            $node->getTemplateName(),
+            $node->getTemplateLine(),
+            null,
+            $messageId,
+        );
+    }
+
+    protected function addError(string $message, Node $node, ?string $messageId = null): bool
+    {
+        return $this->addMessage(
+            Violation::LEVEL_ERROR,
+            $message,
+            $node->getTemplateName(),
+            $node->getTemplateLine(),
+            null,
+            $messageId,
+        );
+    }
+
+    private function addMessage(
+        int $messageType,
+        string $message,
+        ?string $fileName,
+        ?int $line = null,
+        ?int $position = null,
+        ?string $messageId = null
+    ): bool {
+        $fileName ??= $this->filePath;
+
+        if (null === $fileName) {
+            return false;
+        }
+
+        $id = new ViolationId(
+            $this->getShortName(),
+            $messageId ?? ucfirst(strtolower(Violation::getLevelAsString($messageType))),
+            $line,
+            $position,
+        );
+        foreach ($this->ignoredViolations as $ignoredViolation) {
+            if ($ignoredViolation->match($id)) {
+                return false;
+            }
+        }
+
+        $violation = new Violation(
+            $messageType,
+            $message,
+            $fileName,
+            $this->getName(),
+            $id,
+        );
+
+        if (null !== $this->report) {
+            $this->report->addViolation($violation);
+
+            return true;
+        }
+
+        return false;
+    }
+}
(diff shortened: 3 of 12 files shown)


VincentLanglet (commented):
This is an interesting idea. I just have some concern:

1) I consider `customTwigExtension`, `customTokenParser` and `customNodeVisitor` as something internal and not easy to use. So I'd like to reduce the need to use it.

2) The error in the linter will be a fatal one `File is invalid: Filter "trans" is not allowed at line xx` and it will stop the linter for this file so:
- If you have multiple `trans` filter you'll only detect the first one
- You won't have the linter/fixer working for existing rules (as modifying the whitespace)

3) People might ask later for the `ForbiddenFilterNodeVisitor` to replace the filter by something else and I'm not sure it will be easy with the NodeVisitor strategy.

We may decide to ignore the third point, and I think the first point can be sole by introducing something like a NodeVisitorRuleInterface and we would get the NodeVisitorRule from the ruleset to add them automaticaly to the env which would throw custom `ViolationError extends Error` which would be transformed to a Violation. But I don't see a solution for the second point.

WDYT about splitting the PR in half:
- One PR to 'Allow setting custom NodeVisitor on StubbedEnvironment' => I'm ok to merge this. The feature make sens to me ; people might have custom node visitor technically (in the same way they can have token parser)
- One PR for the ForbiddenFilterNodeVisitor => I'm currently unsure about merging this but you can use it on your side.

ruudk:
@VincentLanglet Thanks for the feedback.

For point #2 I already found this out while testing it on my own project. Therefore the solution to just throw an Error in a NodeVisitor is bad.

I refactored the code and I think I have a much better solution now.

This PR is based on https://github.com/VincentLanglet/Twig-CS-Fixer/pull/250 and introduces a `NodeVisitorRuleInterface`. This one is able to add violations to the report. 

What do you think?

Regarding #3, if that's the case, then a new Token based rule should be made. I think with the Node Visitors, it's a great way to do validation based on AST. To block certain things, or to validate certain things.

VincentLanglet (changes requested):

VincentLanglet:
You may need to rebase master to clear the PR diff

src/Rules/NodeVisitorRuleInterface.php
```diff
@@ -0,0 +1,20 @@
...
+namespace TwigCsFixer\Rules;
+
+use TwigCsFixer\Report\Report;
+
+interface NodeVisitorRuleInterface
+{
+    /**
+     * Messages will be added to the given `$report` object.
+     */
+    public function enterFile(Report $report, string $filePath): void;
```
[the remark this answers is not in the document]
ruudk: Funny, this is what I initially did. But I didn't like the fact that `$this->report` could be nullable. 

Also, I feel like when you set it, you should also unset on leaving the file.

That reminds me, I should make sure `$this->violations` is cleared as well. 

Up to you, what you think is better. I will change it.

src/Runner/Linter.php
```diff
@@ -24,10 +26,14 @@ final class Linter
...
 
+    /**
+     * @param list<NodeVisitorInterface> $nodeVisitors
+     */
     public function __construct(
         private Environment $env,
         private TokenizerInterface $tokenizer,
-        ?CacheManagerInterface $cacheManager = null
+        ?CacheManagerInterface $cacheManager = null,
+        private array $nodeVisitors = []
```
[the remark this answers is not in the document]
ruudk:
```php
$ruleset->addRule(new TwigCsFixer\Rules\MyNodeVisitorRule());
```

This `addRule` currently only accepts a `RuleInterface`.

Do you suggest to allow `RuleInterface|NodeVisitorRuleInterface`?

Or was this a typo and did you mean:
```php
$ruleset->addNodeVisitorRule(new TwigCsFixer\Rules\MyNodeVisitorRule());
```

src/Rules/NodeVisitorRuleInterface.php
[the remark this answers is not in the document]
VincentLanglet: In the same way I introduce an AbstractRule to help writing rules, I feel like an AbstractNodeVisitorRule could be helpful and may help to decide what the best strategy.

`AbstractRule::addMessage` might be in a trait (or copy pasted) and should be in `AbstractNodeVisitorRule` too

I'd like to simplify the code to
```
public function enterNode(Node $node, Environment $env): Node
    {
        if (!$node instanceof FilterExpression) {
            return $node;
        }

        if (!\in_array($node->getNode('filter')->getAttribute('value'), $this->forbidden, true)) {
            return $node;
        }

        $this->addError(...);

        return $node;
    }
```
like in classic rules.

By the way, `addMessage` is handling `ignoredViolations` and we should have that too, so the enter file method will need the signature
```
public function enterFile(Report $report, string $filePath, array $ignoredViolations = []): void;
```

Since `private ?Report $report = null;` works fine on AbstractRule, I think it can be the same on `AbstractNodeVisitorRule`.

src/Runner/Linter.php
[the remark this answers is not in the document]
VincentLanglet: > Do you suggest to allow RuleInterface|NodeVisitorRuleInterface?

Yes. I feel like the end-user does not need to know if it's a rule or a nodeVisitorRule.

Looking at `getRules` usage, it's
```
$rules = $ruleset->getRules();
            foreach ($rules as $rule) {
                if ($rule instanceof FixableRuleInterface) {
                    $rule->fixFile($stream, $this, $ignoredViolations);
                }
            }
```
which seems ok and
```
$rules = $ruleset->getRules();
            foreach ($rules as $rule) {
                $rule->lintFile($stream, $report, $ignoredViolations);
            }
```
which could be changed to
```
$rules = $ruleset->getRules();
$nodeVisitorRules = [];
            foreach ($rules as $rule) {
                if ($rule instanceof RuleInterface) {
                    $rule->fixFile($stream, $this, $ignoredViolations);
                } else {
                    $nodeVisitorRules[] = $rule;
                }
            }

if (count($nodeVisitorRules)) {
     // ...
}
```

ruudk:
Thanks for the feedback. I'm going back to the drawing board. Can't fix it right away.

VincentLanglet:
> Thanks for the feedback. I'm going back to the drawing board. Can't fix it right away.

No problem, ping me if some help is needed.

ruudk:
Alright, I made some changes. 

We don't need to have the NodeVisitor's on the Environment. We can just create a new NodeTraverser and traverse the Node's we get after the last parse.

src/Rules/AbstractNodeVisitorRule.php
```diff
@@ -0,0 +1,145 @@
...
+        );
+    }
+
+    protected function addFileWarning(string $message, ?int $line = null, ?string $messageId = null): bool
+    {
+        return $this->addMessage(
+            Violation::LEVEL_WARNING,
+            $message,
+            $this->filePath,
+            null,
```
VincentLanglet, line 81:
```suggestion
            $line,
```

src/Rules/AbstractNodeVisitorRule.php
```diff
@@ -0,0 +1,145 @@
...
+    private function addMessage(
+        int $messageType,
+        string $message,
+        string $fileName,
+        ?int $line = null,
+        ?int $position = null,
+        ?string $messageId = null
+    ): bool {
+        Assert::notNull($this->report);
+        Assert::notNull($this->filePath);
```
VincentLanglet: I don't think we need those assert

ruudk:
How to fix the Psalm error? It doesn't seem to understand Webmozart Assert... Why do we need Psalm and PHPStan? Isn't it easier to pick the most popular one?

ruudk:
Thinking about the naming: NodeVisitorRuleInterface. I feel this is more of an NodeBasedRuleInterface while the others are TokenBasedRuleInterface.

src/Rules/AbstractNodeVisitorRule.php
```diff
@@ -0,0 +1,149 @@
...
+            Violation::LEVEL_WARNING,
+            $message,
+            $this->filePath,
+            null,
+            null,
+            $messageId,
+        );
+    }
+
+    protected function addError(string $message, ?int $line = null, ?string $messageId = null): bool
```
VincentLanglet, line 87: I wonder if we should provide the signature
```
protected function addError(string $message, Node $node, ?string $messageId = null): bool
```
and when we're using
```
$node->getTemplateLine();
$node->getTemplateName();
```

src/Runner/Linter.php
```diff
@@ -122,11 +137,36 @@ public function run(iterable $files, Ruleset $ruleset, ?FixerInterface $fixer =
             }
             restore_error_handler();
 
-            $rules = $ruleset->getRules();
             foreach ($rules as $rule) {
                 $rule->lintFile($stream, $report, $ignoredViolations);
             }
 
+            if ([] !== $nodeVisitorRules) {
+                $this->setErrorHandler($report, $filePath);
```
VincentLanglet, line 145:
```suggestion
```
I think it's not needed here

src/Runner/Linter.php
```diff
@@ -122,11 +137,36 @@ public function run(iterable $files, Ruleset $ruleset, ?FixerInterface $fixer =
...
+                        sprintf('File is invalid: %s', $error->getRawMessage()),
+                        $filePath,
+                        null,
+                        new ViolationId(line: $error->getTemplateLine())
+                    );
+
+                    $report->addViolation($violation);
+                    continue;
+                }
+                restore_error_handler();
```
VincentLanglet, line 161:
```suggestion
```

src/Runner/Linter.php
```diff
@@ -122,11 +137,36 @@ public function run(iterable $files, Ruleset $ruleset, ?FixerInterface $fixer =
...
-            $rules = $ruleset->getRules();
             foreach ($rules as $rule) {
                 $rule->lintFile($stream, $report, $ignoredViolations);
             }
 
+            if ([] !== $nodeVisitorRules) {
+                $this->setErrorHandler($report, $filePath);
+                try {
+                    $twigSource = new Source($content, $filePath);
+                    $node = $this->env->parse($this->env->tokenize($twigSource));
```
VincentLanglet, line 148: If the fixer didn't run, we don't need to parse again the source. We could:
- Write `$node = $this->env->parse($this->env->tokenize($twigSource));` on the `// Validate the file with twig parser.` step
- Reset node ($node = null) only if `if ($newContent !== $content) {` (in the `if (null !== $fixer) {` condition)
- Reparse here the source only if `$node` is null.

src/Runner/Linter.php
```diff
@@ -122,11 +137,36 @@ public function run(iterable $files, Ruleset $ruleset, ?FixerInterface $fixer =
...
+                        new ViolationId(line: $error->getTemplateLine())
+                    );
+
+                    $report->addViolation($violation);
+                    continue;
+                }
+                restore_error_handler();
+
+                foreach ($nodeVisitorRules as $nodeVisitor) {
+                    $nodeVisitor->lintFile($report, $filePath, $ignoredViolations);
```
VincentLanglet, line 164: Since the real `lint` process is done during the `traverse` call, I think we should not call this `lintFile`.

Best I have so far would be `init`/`setFile`/`enterFile` but I dunno, maybe you have something better in mind ?

src/Rules/AbstractNodeVisitorRule.php
```diff
@@ -0,0 +1,149 @@
...
+    private ?Report $report = null;
+
+    private ?string $filePath = null;
+
+    /**
+     * @var list<ViolationId>
+     */
+    private array $ignoredViolations = [];
+
+    public function getName(): string
```
VincentLanglet, line 25: I would say that `getName`, `getShortName` and `addMessage` can be added in a trait used by both Abstract*Rule. But that could be later.

src/Rules/AbstractNodeVisitorRule.php
ruudk: $node->getTemplateName() is nullable.

src/Rules/AbstractNodeVisitorRule.php
ruudk: I agree, but let's do that later.

VincentLanglet:
> How to fix the Psalm error? It doesn't seem to understand Webmozart Assert... Why do we need Psalm and PHPStan? Isn't it easier to pick the most popular one?

You can ignore the failing build, I'll handle them.
I prefer PHPStan over psalm but psalm detect some few things that phpstan didn't. 

Also, this project was for me a fun way to try a lot of dev-tools.

> Thinking about the naming: NodeVisitorRuleInterface. I feel this is more of an NodeBasedRuleInterface while the others are TokenBasedRuleInterface.

Indeed but renaming the RuleInterface to TokenBasedRuleInterface would be a BC break.

Also, you created a `Visitor` folder but we could expect having the rules in the same folder than the interface.
So we could have `Rules/Token/RuleInterface` and `Rules/Node/RuleInterface`, or maybe to avoid the BC break, we keep
`Rules/RuleInterface` and we introduce a folder `Rules/Node` for your rules. In fact it's exactly what they does: in the same way there is rules about punctuation or whitespace, there are now rules about node (?).

ruudk:
Thanks again for the feedback. I made some changes, please have a look 😊

ruudk:
Pushed a fix for the Psalm error. It does indeed find more things.

src/Rules/Node/ForbiddenBlockNodeVisitor.php
```diff
@@ -0,0 +1,40 @@
...
+
+    public function enterNode(Node $node, Environment $env): Node
+    {
+        $blockName = $node->getNodeTag();
+
+        if (null === $blockName) {
+            return $node;
+        }
+
+        if (!\in_array($blockName, $this->forbidden, true)) {
```
VincentLanglet, line 29: If you forbid `trans`, I would expect to forbid `TRANS` too.
I'm not sure, but I think twig is not case sensitive, is it ?

We will need to 
- call `strtolower` on forbidden blocks (on the construct maybe)
- call `strtolower` on blockName

Same for others rules.

src/Rules/Node/ForbiddenFilterNodeVisitor.php
```diff
@@ -0,0 +1,45 @@
...
+use Twig\Node\Node;
+use TwigCsFixer\Rules\AbstractNodeVisitorRule;
+
+final class ForbiddenFilterNodeVisitor extends AbstractNodeVisitorRule
+{
+    /**
+     * @param list<string> $forbidden
+     */
+    public function __construct(
+        private array $forbidden,
```
VincentLanglet, line 18:
```suggestion
        private array $forbiddenFilters,
```
or just `filters` is enough ?

src/Rules/Node/ForbiddenFunctionNodeVisitor.php
```diff
@@ -0,0 +1,45 @@
...
+use Twig\Node\Node;
+use TwigCsFixer\Rules\AbstractNodeVisitorRule;
+
+final class ForbiddenFunctionNodeVisitor extends AbstractNodeVisitorRule
+{
+    /**
+     * @param list<string> $forbidden
+     */
+    public function __construct(
+        private array $forbidden,
```
VincentLanglet, line 18:
```suggestion
        private array $forbiddenFunctions,
```
or just `functions` ?

VincentLanglet (changes requested):
A really great feature.

src/Rules/Node/ForbiddenBlockNodeVisitor.php
```diff
@@ -0,0 +1,40 @@
...
+use Twig\Environment;
+use Twig\Node\Node;
+use TwigCsFixer\Rules\AbstractNodeVisitorRule;
+
+final class ForbiddenBlockNodeVisitor extends AbstractNodeVisitorRule
+{
+    /**
+     * @param list<string> $forbidden
+     */
+    public function __construct(
```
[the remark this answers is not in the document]
ruudk: Done

ruudk:
Applied the feedback again!

VincentLanglet (approved):

VincentLanglet:
Thanks

ruudk:
🥳🥳🥳🥳


Should we revert #250?

VincentLanglet:
> 🥳🥳🥳🥳
> 
> Should we revert #250?

I think it's still a useful feature if someone have custom node visitor (?)

PR #250, opened 2024-07-10
Allow setting custom NodeVisitor on StubbedEnvironment

ruudk:
This is already supported and used by Twig when doing a Parse.

See #249

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -115,7 +115,7 @@ $config->setCacheFile(null);
 return $config;
 ```
 
-## Token parser & Twig Extension
+## Token parser, Twig Extension & Node Visitors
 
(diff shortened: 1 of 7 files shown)


ruudk:
This also needs an entry in the Config. Will work on that.

VincentLanglet (approved):

VincentLanglet:
Seems perfect, can you also add a mention in the
```
## Token parser & Twig Extension
```
sections of the `docs/configuration.md` file (even if the doc is currently far from being perfect ^^')

ruudk:
@VincentLanglet Done, also fixed the code style failure.

VincentLanglet:
Thanks

PR #251, opened 2024-07-10
Refacto

VincentLanglet:


src/Rules/RuleTrait.php
```diff
@@ -0,0 +1,67 @@
...
+            $line,
+            $position,
+        );
+        foreach ($this->ignoredViolations as $ignoredViolation) {
+            if ($ignoredViolation->match($id)) {
+                return false;
+            }
+        }
+
+        $report = $this->report;
```
ruudk, line 52: Instead of this, why not pass this as an argument to the method? Then the trait becomes disconnected from this property having to exist.

VincentLanglet:
(I'll finish tomorrow)

src/Rules/RuleTrait.php
VincentLanglet: Anyway I still need the check for the AbstractRule because the fixer will also use the method but without a report

PR #252, opened 2024-07-11
Improve doc

VincentLanglet:


docs/configuration.md
```diff
@@ -61,6 +61,14 @@ $config->allowNonFixableRules();
...
 ```
 
+## Node-based rules
+
+A few rules are based on the Twig Node and NodeVisitor logic. Because they are different from the default token
+based rules, these rules have some limitations:
+- they cannot be fixable.
+- they can only report the line with the error but not the token position.
+
+Still, these rules can be useful for some static analysis. All of them can be found in `src/Rules/Node` folder.
```
VincentLanglet, line 71: Any recommendation about the doc @ruudk ?

ruudk (approved):
Looks great

llaville:
Hey guys,

WDYT about making Rules documentation in same way as : 

- PHP-CS-Fixer : 1 folder by category and 1 page by Rule.

For example: 
If I took the following rule : `TwigCsFixer\\Rules\\Delimiter\\DelimiterSpacingRule`, then

```
doc/
    delimiter/
        delimiter_spacing_rule.md
```
`CamelCase` to `underscore` format.

This will simplify each rule/page future maintenance and also allow to have a single URL for online documentation in goal of SARIF helpUri feature.

If I told you that, this is because I've almost finish to support [SARIF](https://docs.oasis-open.org/sarif/sarif/v2.1.0/sarif-v2.1.0.html) support for Twig-CS-Fixer, like I did for other projects : https://github.com/llaville/sarif-php-converters/tree/master/docs/converter

I'll then open a Discussion later about SARIF support if you want !

VincentLanglet:
> Hey guys,
> 
> WDYT about making Rules documentation in same way as :
> 
> * PHP-CS-Fixer : 1 folder by category and 1 page by Rule.
> 
> For example: If I took the following rule : `TwigCsFixer\\Rules\\Delimiter\\DelimiterSpacingRule`, then
> 
> ```
> doc/
>     delimiter/
>         delimiter_spacing_rule.md
> ```
> 
> `CamelCase` to `underscore` format.
> 
> This will simplify each rule/page future maintenance and also allow to have a single URL for online documentation in goal of SARIF helpUri feature.
> 
> If I told you that, this is because I've almost finish to support [SARIF](https://docs.oasis-open.org/sarif/sarif/v2.1.0/sarif-v2.1.0.html) support for Twig-CS-Fixer, like I did for other projects : [llaville/sarif-php-converters@`master`/docs/converter](https://github.com/llaville/sarif-php-converters/tree/master/docs/converter?rgh-link-date=2024-07-15T15%3A49%3A09Z)
> 
> I'll then open a Discussion later about SARIF support if you want !

Hi, there is some other docs than rules, so it would require a path like
```
docs/rules/delimiter/delimiter_spacing_rule.md
```

I already something like this in mind (and maybe using github pages to have a website like https://cs.symfony.com/doc/rules/index.html). But all this documentation rework is asking a lot of time, so I would love first having the doc autogenerated (like PHP-CS-Fixer one) if it's possible.

llaville:
Ok, it was an idea to improve this package QA, but there is no hurry !

llaville:
BTW, I've opened a discussion about SARIF support at https://github.com/VincentLanglet/Twig-CS-Fixer/discussions/267

PR #262, opened 2024-07-12
Introduces PHAR build workflow

### LINKED ISSUE #226: Add phar?
Most tools similar to this one (e.g. phpstan, php-cs-fixer) make a phar version available for download.
It would be nice to see the same here.

llaville:
Fixes #226

### DIFF
--- a/.github/workflows/phar.yaml
+++ b/.github/workflows/phar.yaml
@@ -1,0 +1,53 @@
+name: PHAR build
+
+on:
+    push:
+        branches:
+            - main
+    pull_request:
+    workflow_dispatch:
+        
+jobs:
+    phar:
+        name: BOX compile
+
+        runs-on: ${{ matrix.os }}
+
+        strategy:
+            fail-fast: false
+            matrix:
+                os:
+                    - ubuntu-22.04
+                php:
+                    - 8.2
+                tools:
+                    - box:4.6  # available since https://github.com/shivammathur/setup-php/releases/tag/2.27.0
+
+        steps:
+            -   # https://github.com/actions/checkout
+                name: Checkout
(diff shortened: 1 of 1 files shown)


.github/workflows/phar.yaml
```diff
@@ -0,0 +1,53 @@
+name: PHAR build
+
+on:
```
VincentLanglet, line 3: What's the recommended strategy for Phar build ? Should it be on every commit/pull request or only on release ?

llaville: on release it's enough.

VincentLanglet: Then can you update the pr in order to trigger it only on release (and maybe we keep manual dispatch) ? (Or did I misunderstand the current config)

VincentLanglet: I thought about

´´´
on:
    release:
        types: [ published ]
    workflow_dispatch:
´´´

It will automatically run the github actions after I tag a release (or should the phar be generated BEFORE the release ?)

llaville: I cannot help you on this purpose, because personnally, I prefer to use another approach with https://github.com/softprops/action-gh-release combined with https://github.com/miniscruff/changie/ to generate CHANGELOG full and chunk of release version.

For example: https://github.com/llaville/sarif-php-sdk/blob/master/.github/workflows/release.yml

VincentLanglet: The box project seems to be have a github action to upload the phar on release created https://github.com/box-project/box/blob/main/.github/workflows/release.yaml#L83

I'll take a look

PR #269, opened 2024-07-16
Add test

VincentLanglet:


src/Console/Application.php
```diff
@@ -12,25 +12,21 @@ final class Application extends BaseApplication
...
 
-            $version = $installed['versions'][self::PACKAGE_NAME]['pretty_version']
-                ?? $installed['versions'][self::PACKAGE_NAME]['version']
-                ?? 'dev';
-
-            $aliases = $installed['versions'][self::PACKAGE_NAME]['aliases'] ?? [];
-
-            $reference = InstalledVersions::getReference(self::PACKAGE_NAME);
+            $version = $installed['versions'][$package]['pretty_version'] ?? 'dev';
+            $aliases = $installed['versions'][$package]['aliases'] ?? [];
```
VincentLanglet, line 28: @llaville I didn't found a single lib in `vendor/composer/installed.php` which have an alias. I'm thinking about changing 
```
if (null === $reference) {
                return $aliases[0] ?? $version;
            }
```
to
```
if (null === $reference) {
                return $version;
            }
```

What would I lose ?

llaville: When you have only one branch, or if don't specify a composer branch constraint, you won't see the difference.

For example:

`Twig-CS-Fixer dev-main@af956d3`

But when you've project with multiple branches, you are able to see immediatly the difference.

For example: (do it on your repo local copy)


```shell
git checkout -b 3.0
```

```shell
composer update
```

```shell
bin/twig-cs-fixer --version
```

And you should see something like

`Twig-CS-Fixer 3.0.x-dev@af956d3`

VincentLanglet: I tried 
```
git checkout -b 3.0
composer update
bin/twig-cs-fixer --version
```

and i get 
```
Twig-CS-Fixer 3.0.x-dev@af956d3
```
even if I removed the alias code.

VincentLanglet: You can try it on this branch if you want

llaville: This is Composer that do the magic job. Look into `vendor/composer/installed.php` file, 
and you'll get.

```php
        'vincentlanglet/twig-cs-fixer' => array(
            'pretty_version' => '3.0.x-dev',
            'version' => '3.0.9999999.9999999-dev',
            'reference' => 'af956d3e56a8766f66a3e28f9bba25f35a85a943',
            'type' => 'coding-standard',
            'install_path' => __DIR__ . '/../../',
            'aliases' => array(),
            'dev_requirement' => false,
        ),
```

As pretty version has a highter priority in our implementation, you think alias do nothing, but it's not the case

llaville: Without pretty version

```
            $version = null//$installed['versions'][self::PACKAGE_NAME]['pretty_version']
                ?? $installed['versions'][self::PACKAGE_NAME]['version']
                ?? 'dev';
```

you'll get `Twig-CS-Fixer 3.0.9999999.9999999-dev@af956d3`

llaville: Anyway, you'll have the last word. 
The most important is to provide a version that allow to identify this application.

VincentLanglet: I'll try like this, since I cannot test the alias ; I'll see later if it need to be revert.

PR #277, opened 2024-07-20
Introduce ValidConstantFunctionRule

### LINKED ISSUE #248: Verify that constants exist

ruudk:
Fixes #248

This rule can be used to verify correct usage of the `constant` function.

- It requires that the constant (first argument) is passed as string literal. 
- It checks if a `::class` reference was done
- It checks if the constant exists

### DIFF
--- a/src/Rules/Node/ValidConstantFunctionRule.php
+++ b/src/Rules/Node/ValidConstantFunctionRule.php
@@ -1,0 +1,69 @@
+<?php
+
+declare(strict_types=1);
+
+namespace TwigCsFixer\Rules\Node;
+
+use Twig\Environment;
+use Twig\Node\Expression\ConstantExpression;
+use Twig\Node\Expression\FunctionExpression;
+use Twig\Node\Node;
+
+final class ValidConstantFunctionRule extends AbstractNodeRule
+{
+    public function enterNode(Node $node, Environment $env): Node
+    {
+        if (!$node instanceof FunctionExpression) {
+            return $node;
+        }
+
+        $functionName = $node->getAttribute('name');
+        if ('constant' !== $functionName) {
+            return $node;
+        }
+
+        $arguments = $node->getNode('arguments');
+
+        if (1 !== $arguments->count()) {
+            return $node;
+        }
+
+        if (!$arguments->getNode('0') instanceof ConstantExpression) {
+            $this->addError(
+                'Function "constant" expects a constant (string value) as argument.',
+                $node,
+                'ArgumentNotConstant'
+            );
+
+            return $node;
+        }
+
+        $constant = $arguments->getNode('0')->getAttribute('value');
+
+        if (!\is_string($constant)) {
+            return $node;
+        }
+
+        if (\defined($constant)) {
+            return $node;
+        }
+
+        if ('::class' === strtolower(substr($constant, -7))) {
+            $this->addError(
+                sprintf('You cannot use the Twig function "constant()" to access "%s". You could provide an object and call constant("class", $object) or use the class name directly as a string.', $constant),
+                $node,
+                'ClassConstant'
+            );
+
+            return $node;
+        }
+
+        $this->addError(
+            sprintf('Constant "%s" is undefined.', $constant),
+            $node,
+            'ConstantUndefined'
+        );
+
+        return $node;
+    }
+}
--- a/src/Rules/Node/ValidConstantFunctionRule.php
+++ b/src/Rules/Node/ValidConstantFunctionRule.php
@@ -9,6 +9,9 @@ use Twig\Node\Expression\ConstantExpression;
 use Twig\Node\Expression\FunctionExpression;
 use Twig\Node\Node;
 
+/**
+ * Ensures constant function is used on defined constant strings.
+ */
 final class ValidConstantFunctionRule extends AbstractNodeRule
 {
     public function enterNode(Node $node, Environment $env): Node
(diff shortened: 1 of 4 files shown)


src/Rules/Node/ValidConstantFunctionRule.php
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        $arguments = $node->getNode('arguments');
+
+        if (1 !== $arguments->count()) {
+            return $node;
+        }
+
+        if (!$arguments->getNode('0') instanceof ConstantExpression) {
+            $this->addError(
```
VincentLanglet, line 32:
```
{% set var = 'App\\SomeClass::CONSTANT_NAME' %}
{% set foo = constant(var) %}
```
doesn't work ?

We could also have `constant(some_function_generating_a_string())` no ?

ruudk: Indeed, this rule is only allow literal constants. If everything is dynamic then it cannot be verified. The rule is opt in, right? 

If it becomes an issue in the future, we can make a configuration that ignores non literals and does not produce an error. But I don't want it.

ruudk: If you have a function that returns a string why not return the value ? No need to pass to constant function

VincentLanglet: > If you have a function that returns a string why not return the value ? No need to pass to constant function

I agree, but sometimes the `some_function_generating_a_string` method might coming from another (bad) lib an you have no choice.

> The rule is opt in, right?

Sure, but I prefer to "miss" things rather than having "false positive".
I think it's better to drop this error and early return:
- People which does weird thing with `constant` won't have an error for `constant(foo())`.
- People which use constant with string won't see the difference ; I'm pretty sure you never had this error on your codebase don't u ?

> If it becomes an issue in the future, we can make a configuration that ignores non literals and does not produce an error. But I don't want it.

If we do an option, it would be ignored by default IMHO.
But I'm not sure it's worth an option to forbid non-constant-string for the same reason, I don't think you'll have a single non-constant-string in your codebase. Am I wrong ?

VincentLanglet:
I have an issue with this rule when used with the PHAR, all the constant are not defined :/
It's similar with the issue https://github.com/VincentLanglet/Twig-CS-Fixer/issues/226#issuecomment-2241201895

If you have any idea to avoid this @ruudk ? (Or this rule will be only usable with the non-phar version)

ruudk:
Isn't it the same issue as https://github.com/VincentLanglet/Twig-CS-Fixer/issues/226#issuecomment-2241201895?

The PHAR should be able to load project related files. If it isn't able to do that then anything custom won't work. I don't know much about PHAR's and the way to create them to help with this. But i guess it's a solved thing as PHPStan and other tools are also able to do this? Maybe for it to work, the project vendor/autoload.php should be loaded too.

VincentLanglet:
> Isn't it the exact same issue as [#226 (comment)](https://github.com/VincentLanglet/Twig-CS-Fixer/issues/226#issuecomment-2241201895)?

Sure, that's why I asked llaville some help.

> Maybe for it to work, the project vendor/autoload.php should be loaded too.

Yes, I tried 
```
require __DIR__.'/vendor/autoload.php'
```
and got
```
PHP Fatal error:  Declaration of Symfony\Component\Console\Style\OutputStyle::write(Traversable|array|string $messages, bool $newline = false, int $type = self::OUTPUT_NORMAL) must be compatible with Symfony\Component\Console\Output\OutputInterface::write($messages, bool $newline = false, int $options = 0) in vendor/symfony/console/Style/OutputStyle.php on line 49
```

I wasn't able to fix this conflict which I don't have with the non-phar usage.

> I don't know much about PHAR's and the way to create them to help with this

I don't too. That's why I hope @llaville will be able to help me on this.

llaville:
I'm not sure if it's your use case, but when I see that declaration is not compatible, it seems that you try to load two versions of Symfony Console that are differents (and not signature compatible, probably because type hinting).

When I'm in face of such problem, I use https://github.com/bamarni/composer-bin-plugin to handle dependencies conflicts.

VincentLanglet:
> I'm not sure if it's your use case, but when I see that declaration is not compatible, it seems that you try to load two versions of Symfony Console that are differents (and not signature compatible, probably because type hinting).

Yes, indeed.

I'm using the PHAR which has SymfonyConsole in his dependencies and `OutputInterface`.
```
public function write($messages, bool $newline = false, int $options = 0);
```
but I'm trying to use the PHAR on a project which has a newer SymfonyConsole version and `OutputInterface`
```
public function write(Traversable|array|string $messages, bool $newline = false, int $type = self::OUTPUT_NORMAL);
```

The two versions of Symfony Console are loaded because:
- The PHAR uses his dependencies
- I added `vendor/autoload.php` to load all my dependencies (Which should be similar to the boostrap config of your comment https://github.com/VincentLanglet/Twig-CS-Fixer/issues/226#issuecomment-2241237017)

I'll see if box provide something to scope the PHAR https://github.com/box-project/box/blob/main/doc/code-isolation.md#isolating-the-phar

VincentLanglet:
Phar issue seems fixed by https://github.com/VincentLanglet/Twig-CS-Fixer/pull/280

src/Rules/Node/ValidConstantFunctionRule.php
ruudk: FWIW as this is already merged (thanks 😁)
I think [nobody uses `constant` function like that](https://github.com/search?q=lang%3Atwig+%22constant%28%22&type=code).

VincentLanglet: > FWIW as this is already merged. I think [nobody uses `constant` function like that](https://github.com/search?q=lang%3Atwig+%22constant%28%22&type=code).

I agree and think the same. I just prefer to avoid the error just in case.
I bet on the fact that since nobody uses constant with something else than a string, the error isn't needed (since it will never happen on your code base).

> (thanks 😁)

Thanks you for all the good features :)

PR #280, opened 2024-07-20
Add PHAR compactors

VincentLanglet:


llaville:
@VincentLanglet Per my experience , I do not recommend to use the `KevinGH\Box\Compactor\Php` compactor, because it's difficult to found origin of an issue when you've a back trace raised by PHAR execution (file line are changed corresponding to original source code, because phpdoc blocks are stripped). Particular difficult to debug because we have no idea what dependencies are included into the PHAR artifact.
When the `compression` is enabled, the gain is very poor.

VincentLanglet:
> @VincentLanglet Per my experience , I do not recommend to use the `KevinGH\Box\Compactor\Php` compactor, because it's difficult to found origin of an issue when you've a back trace raised by PHAR execution (file line are changed corresponding to original source code, because phpdoc blocks are stripped). Particular difficult to debug because we have no idea what dependencies are included into the PHAR artifact. When the `compression` is enabled, the gain is very poor.

Hi,

I made tests and as soon as I use `KevinGH\\Box\\Compactor\\PhpScoper` it already change the lines. So using the PHP compactor as an extra doesn't add any extra-issues.

PR #282, opened 2024-07-21
Add test for ValidConstantFunctionRule

VincentLanglet:


tests/Rules/Node/ValidConstantFunction/ValidConstantFunctionRuleTest.twig
```diff
@@ -7,4 +7,6 @@
     {{ constant('ThisDoesNotExist::SomeKey') }}
     {{ constant(someVar) }}
     {{ constant('ThisDoesNotExist::class') }}
+    {{ constant(1) }}
```
VincentLanglet, line 10: @ruudk you wrote in your initial PR https://github.com/VincentLanglet/Twig-CS-Fixer/pull/277/commits/bffffcbfe58362034e9cc2c0dc48452e4c876971
```
        $constant = $argument->getAttribute('value');
        if (!\is_string($constant)) {
            return $node;
        }
```

Is there valid case when the value is not a string or should we display an error for this case ?

ruudk: I did this because the return type was mixed according to psalm or PHPStan. Shouldn't be the case I think.

ruudk: So maybe beter to report an error in that case.

src/Rules/Node/ValidConstantFunctionRule.php
```diff
@@ -26,27 +26,39 @@ public function enterNode(Node $node, Environment $env): Node
...
         }
 
         if (\defined($constant)) {
             return $node;
         }
 
+
         if ('::class' === strtolower(substr($constant, -7))) {
             $this->addError(
-                sprintf('You cannot use the Twig function "constant()" to access "%s". You could provide an object and call constant("class", $object) or use the class name directly as a string.', $constant),
```
VincentLanglet, line 49: I'll take any suggestion here @ruudk but I'd like to avoid a too long error message.

So far I always had short message like "Variable should be called foo" or "Expected 2 spaces, found 1". This gave report with all the error messages written on single line every time and I'm not happy with the fact that this error would be certainly written with multiples lines.

Maybe letter something like an additional help text could be written, but so far I'd like a short message.
I try `'You cannot use the function "constant()" to access class names.'` but take any suggestion

ruudk: Fine. This is coming directly from https://github.com/twigphp/Twig/blob/1377f2da4efa020d2ad9b8fb8b722d688ab297ad/src/Extension/CoreExtension.php#L1474

Issue #300, opened 2024-08-16
Rule idea: named argument syntax


smnandre:
## Context

Twig will [soon support](https://github.com/twigphp/Twig/pull/4209) "colons instead of equals for named arguments" 

Before

```twig
{{ include('foo.html.twig', ignore_missing=true) }}
```

After

```twig
{{ include('foo.html.twig', ignore_missing=true) }}

{# or #}
{{ include('foo.html.twig', ignore_missing: true) }}
```

## New rule ? 

A rule could probably be added to standardize, what should be the default? 

Another rule could maybe simply forbid to different characters in the same expression ?

## Other ones ? 

I guess some other rules may require update (i'm thinking about the rule include filter -> function) ?

VincentLanglet:
The first rule which need to be fixed is the HashQuoteRule or it will transform
```
{{ foo(bar: true) }}
```
into
```
{{ foo('bar': true) }}
```
 

I think the best is to change the Tokenizer to tokenize both `=` and `:` into another tokenType like `NAMED_ARGUMENT_OPERATOR`.

Then:
- HashQuoteRule will be unchanged
- A NamedArgumentSpacingRule could be introduced to add/remove space around this operator.
- A NamedArgumentRule could be introduced to replace all of them by `=` or `:` depending of the preferred value.

> what should be the default?

Not sure we can use `=` by default since it's only supported on last twig version...

smnandre:
> Not sure we can use = by default since it's only supported on last twig version...

You mean `:` right ?

VincentLanglet:
> > Not sure we can use = by default since it's only supported on last twig version...
> 
> You mean `:` right ?

yes, edited

VincentLanglet:
Or we could promote the new `:` syntax but add an early return if it's not supported by Twig (based on the installed version ?)

smnandre:
Let's wait to see how this goes. Personnaly i'm all in to switch to the new syntax, but there may be drawbacks i'm not seeing yet :)

VincentLanglet:
I think I implemented all you wanted in https://github.com/VincentLanglet/Twig-CS-Fixer/pull/302.

I'll just wait for the twig 3.12 release.

smnandre:
>  all you wanted

But i was just conversing, i did not "want" anything 😅 

But that's super great! And i think this will be very useful for a lot of people when 3.12 is released :) 

👏

smnandre:
Well, we have our answer


[...]

PR #302, opened 2024-08-17
Introduce NAMED_ARGUMENT_SEPARATOR_TYPE and related rules.

VincentLanglet:
Closes https://github.com/VincentLanglet/Twig-CS-Fixer/issues/300


src/Token/Token.php
```diff
@@ -41,6 +41,7 @@ final class Token
     public const COMMENT_TAB_TYPE = 'COMMENT_TAB_TYPE';
     public const COMMENT_EOL_TYPE = 'COMMENT_EOL_TYPE';
     public const COMMENT_END_TYPE = 'COMMENT_END_TYPE';
+    public const NAMED_ARGUMENT_OPERATOR_TYPE = 'NAMED_ARGUMENT_OPERATOR_TYPE';
```
VincentLanglet, line 44: I'm open to better/shorter name @smnandre

src/Token/Tokenizer.php
```diff
@@ -587,6 +590,17 @@ private function lexOperator(string $operator): void
         } elseif (':' === $operator && $this->isInTernary()) {
             $bracket = array_pop($this->bracketsAndTernary);
             $this->pushToken(Token::OPERATOR_TYPE, $operator, $bracket);
+        } elseif ('=' === $operator) {
+            $bracket = end($this->bracketsAndTernary);
+            if (false !== $bracket && '(' === $bracket->getValue()) {
```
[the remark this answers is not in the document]
smnandre: in named arguments for functions and filters (so you may have multiple ones)

src/Token/Token.php
smnandre: NAMED_ARGUMENT_OPERATOR feels not bad (maybe without the "type") ?

src/Token/Tokenizer.php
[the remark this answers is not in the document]
VincentLanglet: I just thought about default value for macro.
`{% macro input(name='foo') %}{{ name }}{% endmacro %}`

This one need to be handled differently since it cannot be changed to `:`.

> in named arguments for functions and filters (so you may have multiple ones)

functions and filters (and tests) should work the same (and mutliple named operator too).

src/Token/Token.php
VincentLanglet: > (maybe without the "type") ?

So far I suffixed all the tokens by TYPE (like twig did)... So I kinda have to continue

smnandre: Ho sorry :) I'm switching between a few projects i'm losing my mind haha

src/Rules/Operator/NamedArgumentOperatorRule.php
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Ensures named arguments use `:` syntax instead of `=` (For `twig/twig >= 3.12.0`).
+ */
+final class NamedArgumentOperatorRule extends AbstractFixableRule
+{
+    public function __construct()
+    {
+        if (!InstalledVersions::satisfies(new VersionParser(), 'twig/twig', '>=3.12.0')) {
+            throw new \InvalidArgumentException('Named argument with semi colons requires twig/twig >= 3.12.0');
+        }
```
smnandre, line 22: _**If**_ you want to, you can use the Twig Environment constants for this.

VincentLanglet: Not sure it can be trusted, when we see mistake like https://github.com/twigphp/Twig/blob/v3.11.0/src/Environment.php#L48

PR #310, opened 2024-08-30
Fix deprecation in ForbiddenFilterRule.php

ruudk:
> Deprecate the `filter` node of `FilterExpression`

Needed for Twig 3.12.0

See https://github.com/twigphp/Twig/pull/4184

Do we need to bump the minimum required Twig version?

### DIFF
--- a/src/Rules/Node/ForbiddenFilterRule.php
+++ b/src/Rules/Node/ForbiddenFilterRule.php
@@ -35,7 +35,9 @@ final class ForbiddenFilterRule extends AbstractNodeRule implements Configurable
             return $node;
         }
 
-        $filterName = $node->getNode('filter')->getAttribute('value');
+        $filterName = $node->hasAttribute('name')
+            ? $node->getAttribute('name')
+            : $node->getNode('filter')->getAttribute('value'); // BC for twig/twig < 3.12
         if (!\in_array($filterName, $this->filters, true)) {
             return $node;
         }


VincentLanglet:
> > Deprecate the `filter` node of `FilterExpression`
> 
> Needed for Twig 3.12.0

Shouldn't the same be done for FunctionExpression too ?
https://github.com/VincentLanglet/Twig-CS-Fixer/blob/main/src/Rules/Node/ForbiddenFunctionRule.php#L38
https://github.com/VincentLanglet/Twig-CS-Fixer/blob/main/src/Rules/Node/ValidConstantFunctionRule.php#L23

> Do we need to bump the minimum required Twig version?

No, I add to extends the lower bound support of twig/twig for drupal and some others CMS so i'll keep it.
https://github.com/VincentLanglet/Twig-CS-Fixer/pull/306

We need to support both, with something like
```
$node->hasAttribute('twig_callable')
  ? $node->getAttribute('twig_callable')->getName() // twig/twig >= 3.12
  : $node->getNode('filter')->getAttribute('value');
```

VincentLanglet:
> We need to support both, with something like
> 
> ```
> $node->hasAttribute('twig_callable')
>   ? $node->getAttribute('twig_callable')->getName() // twig/twig >= 3.12
>   : $node->getNode('filter')->getAttribute('value');
> ```

For the phpstan issue it will be
```
if ($node->hasAttribute('twig_callable')) {
    $twigCallable = $node->getAttribute('twig_callable');
    Assert::instanceof($twigCallable, TwigFilter::class);
    $filterName = $twigCallable->getName();
} else {
    // BC for twig/twig < 3.12
    $filterName = $node->getNode('filter')->getAttribute('value');
}
```

ruudk:
I guess not, the name attribute was only added in the latest version. Force pushed for another try. This looks better?

VincentLanglet:
You need https://github.com/VincentLanglet/Twig-CS-Fixer/pull/311 and to rebase after

VincentLanglet:
> I guess not, the name attribute was only added in the latest version. Force pushed for another try. This looks better?

What about FunctionExpression ? Should it be fixed too ?

Also @ruudk , I don't have any deprecation when running my tests, I would be interested if you know how to have them displayed by PHPUnit to confirm
- I have deprecation before
- And I have no deprecation after your PR

ruudk:
Good idea. Let me try to enable that first, so you are aware of these deprecations automatically.

ruudk:
We need to install the Symfony PHPUnit Bridge in order to get the deprecations to fail the test.

But this also immediately will give you a problem for the future: the Symfony PHPUnit Bridge is not compatible with PHPUnit 10 / 11. See https://github.com/symfony/symfony/issues/49069.

But luckily, PHPUnit 11.2 also has this functionality now. But for Symfony (not related to you I guess) there is another issue: https://github.com/symfony/symfony/issues/53812 

So what to do? Use the PHPUnit Bridge for now and enable the deprecations helper?

VincentLanglet:
I will wait Phpunit 11.2 then

ruudk:
> What about FunctionExpression ? Should it be fixed too ?

That already uses `->getAttribute('name')`

VincentLanglet:
> > What about FunctionExpression ? Should it be fixed too ?
> 
> That already uses `->getAttribute('name')`

Ok, can you just rebase main @ruudk

VincentLanglet (approved):

VincentLanglet:
Thanks @ruudk, Do you have other findings related to the 3.12 version or should i release soon ?

ruudk:
Not that I know. It worked when we upgraded this morning. But since I fixed a few deprecations in our project I went to see if there were any deprecations here. So I guess all just works fine ☺️

PR #314, opened 2024-09-12
Allow variables containing dollar signs

### LINKED ISSUE #313: Allow variables starting with dollar sign ($)
### Context

We'd like to use Twig-CS-Fixer for our [Shopware](https://github.com/shopware/shopware) projects (custom plugins / Symfony bundles).

Shopware uses Twig both directly from PHP, but also from JavaScript via [twigjs/twig.js](https://github.com/twigjs/twig.js).

Since twig.js implements the same syntax as the reference Twig implementation, Twig-CS-Fixer mostly works fine, but we ran into a little problem:

Shopware uses a Variable / Function called `$tc` for translations (e.g. `$tc('hello')`). Note the dollar sign at the start.

While this is supported by twig.js, the reference Twig implementation and Twig-CS-Fixer do not support variables starting with a dollar sign.

For compatibility it would be nice for Twig-CS-Fixer to also allow the dollar sign at the start of variables.

### Expected behavior

Variables starting with a dollar sign can be parsed.

Write here.

### Actual behavior

Tokenizing fails.

```
 KO .../example.html.twig
 ------- --------------------------------------------------------------------- 
  FATAL   >>   | Unable to tokenize file: Unexpected character "$" at line 3.  
 ------- --------------------------------------------------------------------- 
```

nussjustin-hmmh:
Fixes #313

### DIFF
--- a/src/Token/Tokenizer.php
+++ b/src/Token/Tokenizer.php
@@ -11,7 +11,7 @@ use TwigCsFixer\Report\ViolationId;
 use Webmozart\Assert\Assert;
 
 /**
- * An override of Twig\Lexer to add whitespace and new line detection.
+ * An override of Twig\Lexer to add whitespace and new line detection and allow variable names containing dollar signs.
  */
 final class Tokenizer implements TokenizerInterface
 {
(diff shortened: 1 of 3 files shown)


nussjustin-hmmh:
Closed. See [my comment in #313](https://github.com/VincentLanglet/Twig-CS-Fixer/issues/313#issuecomment-2346120835)

PR #317, opened 2024-10-10
Move AbstractRuleTestCase and TestHelper to src

### LINKED ISSUE #316: AbstractRuleTestCase is not part of the Composer package
I found it incredibly easy to write a custom rule for my project. (Thank you!) However, I found it challenging to write PHPUnit tests for that custom rule. 

I think `\TwigCsFixer\Tests\Rules\AbstractRuleTestCase` (and maybe other test cases) should be moved from `tests/` to `src/` so that it is available when installed as a Composer dependency.

Unless there's a way to do this that I'm missing. Thoughts?

### Expected behavior

I can extend `\TwigCsFixer\Tests\Rules\AbstractRuleTestCase` in my own codebase.

### Actual behavior

`\TwigCsFixer\Tests\Rules\AbstractRuleTestCase` does not exist.

zacharylund:
Fixes #316

### DIFF
--- a/composer-dependency-analyser.php
+++ b/composer-dependency-analyser.php
@@ -10,6 +10,11 @@ $config = new Configuration();
 return $config
     ->addPathToScan(__DIR__.'/bin/twig-cs-fixer', isDev: false)
     ->ignoreErrorsOnPackageAndPath(
+        'phpunit/phpunit',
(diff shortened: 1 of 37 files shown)


VincentLanglet:
Thanks, @zacharylund 

Can you try dev-main on your side and tell me if you have all the class you need ?

PR #323, opened 2024-11-05
Support type block

### LINKED ISSUE #322: Support optionality in new `{% types %}` tag
The new `types` tag supports optional variables like this:

```twig
{% types {
    name?: 'string',
} %}
```

But this is currently detected as a problem:
```
 KO ./tests/EndToEnd/Types/optionality.twig
 ------- -----------------------------------------------------
  ERROR   6    | {% types {
          7    |     name?: 'string',
          >>   | Expecting 1 whitespace before "?:"; found 0.
          8    | } %}
 ------- -----------------------------------------------------
```

And fixed like this:
```twig
{% types {
    name ?: 'string|null',
} %}
```

which is not valid.

VincentLanglet:
Closes #322


ruudk (approved):
Great, thanks for the fast fix!

PR #324, opened 2024-11-05
Update .gitattributes

ruudk:
Screenshot 2024-11-05 at 11 40 26@2x

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,14 +1,12 @@
 .* export-ignore
-patches export-ignore
+docs export-ignore
 tests export-ignore
 infection.json export-ignore
 Makefile export-ignore
-composer-require-checker.json export-ignore
+box.json.dist export-ignore
+composer-dependency-analyser.php export-ignore
 phpstan.neon export-ignore
 phpunit.xml.dist export-ignore
 psalm.xml export-ignore
 rector.php export-ignore
-box.json.dist export-ignore
-composer-dependency-analyser.php export-ignore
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -8,3 +8,7 @@ phpstan.neon export-ignore
 phpunit.xml.dist export-ignore
 psalm.xml export-ignore
 rector.php export-ignore
+box.json.dist export-ignore
+composer-dependency-analyser.php export-ignore
+scoper.inc.php export-ignore
+docs export-ignore
```
VincentLanglet, line 14: Isnt it useful to keep docs ?
Especially because there are listed in the readme

ruudk: I looked at Twig and they don't it. 

There might be a case where this is useful sometimes. But we should be conservative with what we archive and download every time on CI ;)

VincentLanglet: Ok, can you update it too
```
.* export-ignore
patches export-ignore
docs export-ignore
tests export-ignore
infection.json export-ignore
Makefile export-ignore
box.json.dist export-ignore
composer-dependency-analyser.php export-ignore
phpstan.neon export-ignore
phpunit.xml.dist export-ignore
psalm.xml export-ignore
rector.php export-ignore
scoper.inc.php export-ignore
```
this way it's in the same order than in the repo, and

`composer-require-checker.json export-ignore` can be removed

ruudk: Done

VincentLanglet (approved):

PR #373, opened 2025-05-19
Try reproducer

VincentLanglet:


tests/Rules/Literal/CompactHashRule/CompactHashRuleTest.php
```diff
@@ -42,4 +42,11 @@ public function testRuleWithCompact(): void
             'CompactHash.Error:20:5' => 'Hash key "product" should be implicit.',
         ], fixedFilePath: __DIR__.'/CompactHashRuleTest.compact.fixed.twig');
     }
+
+    public function testJDreesen(): void
+    {
+        $this->checkRule(new CompactHashRule(true), [
```
jdreesen, line 48: It happens when the `PunctuationSpacingRule` runs before `CompactHashRule`.

In this case, `\TwigCsFixer\Runner\Fixer::$fixedTokens` already contains the token, so https://github.com/VincentLanglet/Twig-CS-Fixer/blob/01afc656e5599abfc166aead6ed7d145474eaef5/src/Rules/Literal/CompactHashRule.php#L118 does nothing.

```suggestion
        $this->checkRule(new PunctuationSpacingRule(), new CompactHashRule(true), [
```

PR #379, opened 2025-06-18
Update doc about CompactHashRule

VincentLanglet:


docs/rules.md
```diff
@@ -172,6 +177,7 @@ new TwigCsFixer\Rules\Whitespace\IndentRule(3);
 - Twig
 - BlankEOFRule
 - BlockNameSpacingRule
+- CompactHashRule
```
ruudk, line 180: I wonder why the default is to have this turned off, instead of on?

stof: the main reason why Symfony uses explicit keys is probably because the compact notation was added later (and for some time, Symfony wanted the WebProfilerBundle to stay compatible with older Twig versions.

Note that the renaming issue is only the case when using naive search and replace. When using the refactoring tool of the IDE, it should be capable to expand it properly if it supports this notation..

PR #388, opened 2025-08-03
docs: use h3 in standards list in rules.md

smnandre:
I'm suggesting to use titles here, because it allows to link directly to a given standard set of rules...

Also this will increase visibility in the Table of Content (on GitHub or ... in the sidebar 👼 )

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -163,7 +163,8 @@ new TwigCsFixer\Rules\Whitespace\IndentRule(3);
 
 ## Standards
 
-**Twig**:
+### Twig Rules
+
 - DelimiterSpacingRule
 - NamedArgumentNameRule
 - NamedArgumentSeparatorRule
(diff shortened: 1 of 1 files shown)


docs/rules.md
```diff
@@ -163,7 +163,8 @@ new TwigCsFixer\Rules\Whitespace\IndentRule(3);
 
 ## Standards
 
-**Twig**:
+### Twig Rules
```
VincentLanglet, line 166:
```suggestion
### Twig
```
for others standards you kept
- TwigCsFixer
- Symfony

without adding `Rules`, so let's keep consistency no ?

smnandre: My bad, thank you!

VincentLanglet (approved):

VincentLanglet:
Thanks

PR #420, opened 2025-12-17
chore(deps): allow webmozart/assert v2

trearcul:


trearcul:
This also needs to wait for infection/infection to allow webmozart/assert v2; otherwise v2 won’t be installed in dev environments.

VincentLanglet:
infection does support webmozart 2 now, so I merge this, thanks.

PR #443, opened 2026-03-27
Add StrictComparisonOperatorRule

HSAIG:
# Add StrictComparisonOperatorRule

## Summary

This PR introduces a new optional fixable rule: `StrictComparisonOperatorRule`.

It replaces Twig strict comparison tests:

* `is same as(...)` → `===`
* `is not same as(...)` → `!==`

## Motivation

Twig 3.23 introduced `===` and `!==` as strict comparison operators.

Using these operators improves readability and aligns Twig syntax with PHP and JavaScript.

## Details

* The rule should be optional
* It only applies to explicit `same as` / `not same as` cases
* It is intended for projects using Twig >= 3.23

## Example

**Before**

```twig
{% if foo is same as(bar) %}
{% endif %}
```

**After**

```twig
{% if foo === bar %}
{% endif %}
```


src/Rules/Operator/StrictComparisonOperatorRule.php
```diff
@@ -0,0 +1,108 @@
...
+    {
+        $token = $tokens->get($tokenIndex);
+
+        if (!$token->isMatching(Token::OPERATOR_TYPE, ['is', 'is not'])) {
+            return;
+        }
+
+        $isNegated = $token->isMatching(Token::OPERATOR_TYPE, 'is not');
+
+        $sameAsIndex = $tokens->findNext(Token::WHITESPACE_TOKENS, $tokenIndex + 1, exclude: true);
```
VincentLanglet: Exclude EMPTY_TOKENS rather than WHITESPACE.

Cause here you don't support someone with an indent in the middle of the code or a new line. (And add test with this case)

VincentLanglet (changes requested):

VincentLanglet:
Hi @HSAIG, I had to revert this PR while playing with it because an issue about operator priority.

See https://twig.symfony.com/play?data=eyJ0ZW1wbGF0ZXMiOltbImluZGV4LnR3aWciLCJGaXJzdDoge3sgYSBpcyBzYW1lIGFzIGIgKyAxfX1cblxuU2Vjb25kOiB7eyBhID09IGIgKyAxfX0iXV0sImNvbnRleHQiOnsiYSI6IjEiLCJiIjoiMSJ9LCJ2ZXJzaW9uIjoiMy4yMS4xIiwib3B0aW9ucyI6eyJzdHJpY3RfdmFyaWFibGVzIjp0cnVlLCJjaGFyc2V0IjoiVVRGLTgiLCJhdXRvZXNjYXBlIjoiIn19
or https://twig.symfony.com/play?data=eyJ0ZW1wbGF0ZXMiOltbImluZGV4LnR3aWciLCJGaXJzdDoge3sgYSBpcyBzYW1lIGFzKGIpICsgMX19XG5cblNlY29uZDoge3sgYSA9PSBiICsgMX19Il1dLCJjb250ZXh0Ijp7ImEiOiIxIiwiYiI6IjEifSwidmVyc2lvbiI6IjMuMjEuMSIsIm9wdGlvbnMiOnsic3RyaWN0X3ZhcmlhYmxlcyI6dHJ1ZSwiY2hhcnNldCI6IlVURi04IiwiYXV0b2VzY2FwZSI6IiJ9fQ%3D%3D
(I show the issue with `==`, I assumed it's the same with `===` but the playground is in 3.21.1) 

`same as` and `===` (or `==` here) does not have the same priority so I assume it cannot (always ?) be automatically fixed...

BTW I made some improvement in https://github.com/VincentLanglet/Twig-CS-Fixer/pull/445 like skipping the rule on twig version prior to 3.23.

HSAIG:
:( Sadly I did not think about that...
And you are right, the precedence is the same for `==` and `===` (https://twig.symfony.com/doc/3.x/operators_precedence.html) which is 20, whereas for `is` it is 100.

Anyway, thanks for your time reviewing, testing and improving my code ^^

PR #457, opened 2026-06-27
Add named arguments option to IncludeFunctionRule

nlemoine:
Salut!

Just thought it would be nice to have named arguments when applying the `IncludeFunctionRule` rule. Defaults to `false` to avoid breaking changes.

Named args usually provide better readability than long list of booleans. 

Happy to get your feedback on this.

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -37,9 +37,10 @@
   Ensures that hash key are not unnecessarily quoted. Options are:
     - `useQuote`: hash key must be preferred quoted (default false).
 
-- **IncludeFunctionRule**:
+- **IncludeFunctionRule** (Configurable):
 
-  Ensures that include function is used instead of include tag.
+  Ensures that include function is used instead of include tag. Options are:
+    - `namedArguments`: use named arguments for `with_context` and `ignore_missing` instead of positional ones (default false).
 
 - **IndentRule** (Configurable):
 
--- a/src/Rules/Function/IncludeFunctionRule.php
+++ b/src/Rules/Function/IncludeFunctionRule.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace TwigCsFixer\Rules\Function;
 
 use TwigCsFixer\Rules\AbstractFixableRule;
+use TwigCsFixer\Rules\ConfigurableRuleInterface;
 use TwigCsFixer\Token\Token;
 use TwigCsFixer\Token\Tokens;
 use Webmozart\Assert\Assert;
@@ -12,8 +13,19 @@ use Webmozart\Assert\Assert;
 /**
  * Ensures that the include function is used instead of include tag.
  */
-final class IncludeFunctionRule extends AbstractFixableRule
+final class IncludeFunctionRule extends AbstractFixableRule implements ConfigurableRuleInterface
 {
+    public function __construct(private readonly bool $namedArguments = false)
+    {
+    }
+
+    public function getConfiguration(): array
+    {
+        return [
+            'namedArguments' => $this->namedArguments,
+        ];
+    }
+
     protected function process(int $tokenIndex, Tokens $tokens): void
     {
         $token = $tokens->get($tokenIndex);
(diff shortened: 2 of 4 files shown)


VincentLanglet (commented):
I feel like the option (which might need a better name then) should only use named argument if necessary

For instance
```
{{ include(['template_a.html', 'template_b.html'], {'foo': 'bar'}, with_context=false, ignore_missing=true) }}
```
could stay
```
{{ include(['template_a.html', 'template_b.html'], {'foo': 'bar'}, false, true) }}
```

nlemoine:
Option naming apart, I'm not getting it. The example you posted is the actual state of the current fix. My proposal is the exact opposite (named argument everywhere). 

```twig
{{ include(['template_a.html', 'template_b.html'], {'foo': 'bar'}, false, true) }}
```

☝️ This is the exact syntax I want to fix with the PR rule option ^^

VincentLanglet:
> ```
> {{ include(['template_a.html', 'template_b.html'], {'foo': 'bar'}, false, true) }}
> ```
> 
> ☝️ This is the exact syntax I want to fix with the PR rule option ^^

Forcing namedArgument for with_context/ignore_missing in this syntax is 100% opinionated and won't be accepted.
Why with_context/ignore_missing and not other param too ?

But asking to fix 
```
{% include 'template.html' ignore missing %}
```
into
```
{{ include('template.html', ignore_missing=true) }}
```
rather than the existing
```
{{ include('template.html', [], true, true) }}
```
would make sens to me cause it saves some params.

nlemoine:
> Forcing namedArgument for with_context/ignore_missing in this syntax is 100% opinionated and won't be accepted.

Hence why it's both **optional** and **off** by default, it's a behavior user deliberately toggle on. It's more about DX than saving params here. I'm not going to list the countless blog posts that make the named args case (TL; DR; `doSomething(true, false, true, true)` is arguably terrible DX and it applies here).

But if you feel the rule doesn't belong to Twig CS Fixer core, it's ok :) I'll close the PR and move the rule to my project or another package.

VincentLanglet:
> Hence why it's both **optional** and **off** by default, it's a behavior user deliberately toggle on. It's more about DX than saving params here. I'm not going to list the countless blog posts that make the named args case (TL; DR; `doSomething(true, false, true, true)` is arguably terrible DX and it applies here).

Understand the issue as a maintainer.
- What if someone ask to have namedParameter for every param and not only the two you asked for ?
- What if someone ask to have namedParameter only when necessary to avoid extra param ?

How will I introduce this later without having tons of option ?
Would it be better to introduce constant and then having something like an option `namedArguments` with the choices:
'none', 'booleans', 'minimal', 'all' ?

nlemoine:
> What if someone ask to have namedParameter for every param and not only the two you asked for ?

That's the rule I was about to propose. Make it live as a standalone rule, not tied to `IncludeFunctionRule`. Would you be open to that?

> What if someone ask to have namedParameter only when necessary to avoid extra param ?

Honestly, thinking as a maintainer, I would simply not go that way. And it all resides in the "only when necessary": it's more machinery (and consequently more maintenance) and possibly very subjective. Same as 'none', 'booleans', 'minimal', 'all'.

IMHO, that rule isn't intended to save params. They no real point in that. It's about readability and maintainability:
- It restores the semantics the tag syntax already had: `{% include 'foo.html' ignore missing %}` -> `{{ include('foo.html', ignore_missing = true) }}`
- It aligns with where Twig and PHP are already going
- It makes args explicit, especially valuable in Twig where IDE support is sparse

So either you enable or disable the rule. 

Let me know if that approach would work for you.

VincentLanglet:
> > What if someone ask to have namedParameter for every param and not only the two you asked for ?
> 
> That's the rule I was about to propose. Make it live as a standalone rule, not tied to `IncludeFunctionRule`. Would you be open to that?

Sure I would be open to that, but I'm not sure if it's really possible because it would require to know/guess the name of the params...

> > What if someone ask to have namedParameter only when necessary to avoid extra param ?
> 
> Honestly, thinking as a maintainer, I would simply not go that way. And it all resides in the "only when necessary": it's more machinery (and consequently more maintenance) and possibly very subjective. Same as 'none', 'booleans', 'minimal', 'all'.

The fact is subjective is my main pain point. 
Personally I wouldn't be interested by the named parameter option you provide but by an option (the `'boolean'` version)  but would prefer to use the 'minimal' version. Given the fact we're already have two different point of view (with only two interlocutors in this discussion), we can assume their will be more and more subjective opinion about this.

Given that, I don't feel confident merging an option `namedParameters : true/false` because the API won't be easy to change in the futur without a BC break and a new major version. The `'none', 'booleans', 'minimal', 'all'.` option is the best way to provide a more flexible API in the future.

I'm okay if we just implement 2-3 options like `none, booleans, all` or `none, boolean, minimal` for now.

PR #461, opened 2026-08-19
Support Twig 3.29 documentation comments in DelimiterSpacingRule

Kocal:
Twig 3.29 ([twigphp/Twig#4871](https://github.com/twigphp/Twig/pull/4871)) adds documentation comments: `{## text #}`, and the whitespace-trim variants `{##- text -#}` / `{##~ text ~#}`. `DelimiterSpacingRule` currently flags them with `Expecting 1 whitespace after "{#"`.

This makes the rule accept a `#` doc marker directly after the `{#` opening, while still enforcing a single space after the `##` marker and a space before the closing `#}`. The empty comment `{##}` keeps its meaning.

Tests cover the rule and the tokenizer, and `docs/rules.md` is updated.


VincentLanglet:
Thanks for the Pr

I'll check it after my vacations (end of the month). Tonsee if it's the right fix or i should update the parser.

We also need to check the syntax with type works fine:
´´´
{% types {
    ## The unique identifier of the article
    id: 'string',

    ## Whether the article should be highlighted
    featured?: 'boolean',
} %}
´´´

Kocal:
No rush, enjoy the vacation!

I checked `{% types %}` (and `{% props %}` too, it's the same thing): inline `##` works fine, no false positives. 

This PR only touches the `{## #}` form, and inline `##` never reaches the rule, so there's nothing to fix there. I added a test to cover it.

VincentLanglet:
Hi @Kocal.

I preferred to fix this on the Tokenizer side by introducing new tokens.
Could you confirm me that https://github.com/VincentLanglet/Twig-CS-Fixer/pull/462 works fine for u ?

Kocal:
Hey @VincentLanglet , I confirm that your PR works fine as well
I'm closing this one

PR #462, opened 2026-08-30
Tokenize doc comment

VincentLanglet:


Kocal:
Since you tokenized the comments, does it means that in the future, we will be able to automatically fixes the following code:
```twig
{% props
      ## string The URL the form submits to.
    href,
   ## string|null The CSRF token ID used to generate the hidden token field; when null, no token is added.
    csrfTokenId = null,
    ## string The HTTP method used to submit the form; when not `POST`, a hidden `_method` field is added for method spoofing.
    method = 'POST',
%}
```

to:
```twig
{% props
    ## string The URL the form submits to.
    href,
    ## string|null The CSRF token ID used to generate the hidden token field; when null, no token is added.
    csrfTokenId = null,
    ## string The HTTP method used to submit the form; when not `POST`, a hidden `_method` field is added for method spoofing.
    method = 'POST',
%}
```

?

VincentLanglet:
> Since you tokenized the comments, does it means that in the future, we will be able to automatically fixes the following code:
> 
> ```
> {% props
>       ## string The URL the form submits to.
>     href,
>    ## string|null The CSRF token ID used to generate the hidden token field; when null, no token is added.
>     csrfTokenId = null,
>     ## string The HTTP method used to submit the form; when not `POST`, a hidden `_method` field is added for method spoofing.
>     method = 'POST',
> %}
> ```
> 
> to:
> 
> ```
> {% props
>     ## string The URL the form submits to.
>     href,
>     ## string|null The CSRF token ID used to generate the hidden token field; when null, no token is added.
>     csrfTokenId = null,
>     ## string The HTTP method used to submit the form; when not `POST`, a hidden `_method` field is added for method spoofing.
>     method = 'POST',
> %}
> ```
> 
> ?

This was already technically achievable since this was tokenized as INLINE_COMMENT.
Now I differentiate inline comment and inline doc comment in case someone wants different rule.

For instance a rule might be that a doc comment starting with `{##` should end with `##}` and not `#}`.

The difficulty in the rule you ask is about the indentation.
What is the right indentation to use ?
In your exemple we could consider that the doc comment has to use the same indentation than the next line.

So it will fix 
> ```
> {% props
>       ## string The URL the form submits to.
>     href,
>         ## string|null The CSRF token ID used to generate the hidden token field; when null, no token is added.
>          csrfTokenId = null,
>     ## string The HTTP method used to submit the form; when not `POST`, a hidden `_method` field is added for method spoofing.
>     method = 'POST',
> %}
> ```
> 
> to:
> 
> ```
> {% props
>     ## string The URL the form submits to.
>     href,
>          ## string|null The CSRF token ID used to generate the hidden token field; when null, no token is added.
>          csrfTokenId = null,
>     ## string The HTTP method used to submit the form; when not `POST`, a hidden `_method` field is added for method spoofing.
>     method = 'POST',
> %}
> ```

easily. (Look for inline_doc_comment, look for the indent of next line and compare)

Another strategy could be to consider the right indentation is the one of the props block + 4.
But I dunno if it could have bad impact.

Anyway, it's possible to do thing. The only thing which is too complicated to indent is html (cf some existing closed issues)

